PR #6, opened 2025-07-22
Point demo to prod backend in prod

its-everdred:
## Problem

The verbs frontend is hardcoded to localhost

## Solution

Make dynamic, set defaults, and allow override

### DIFF
--- a/packages/demo/frontend/src/envVars.ts
+++ b/packages/demo/frontend/src/envVars.ts
@@ -1,0 +1,20 @@
+import { z } from 'zod'
+
+const envVarSchema = z.object({
+  VITE_VERBS_API_URL: z
+    .string()
+    .url()
+    .default(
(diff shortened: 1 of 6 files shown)


packages/demo/frontend/src/envVars.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from 'zod'
+
+const envVarSchema = z.object({
+  VITE_VERBS_API_URL: z
+    .string()
+    .url()
+    .default(
+      import.meta.env.MODE === 'production'
```
its-everdred, line 8: 🎨 `MODE` is provided by vite:
`pnpm run dev` sets `development`
`pnpm preview` sets `production`

its-everdred (commented):
Merging since this is blocking, happy to pull any feedback into the next PR

PR #28, opened 2025-07-25
feat: Add balance fetching infrastructure

tremarkley:
Part of https://github.com/ethereum-optimism/verbs/issues/27

## Overview
This PR introduces a comprehensive token balance fetching system for the Verbs SDK, including new services for chain management and token registry, along with enhanced token functionality.

### `ChainManager` Service
- New ChainManager class for managing blockchain connections and public clients
- Supports multiple RPC endpoints with client creation
- Added test suite (`ChainManager.spec.ts`)

### `TokenRegistry` Service
- New TokenRegistry class for managing token instances
- Simple registration and retrieval of tokens by symbol
- Foundation for extensible token ecosystem
- Added test suite (`TokenRegistry.spec.ts`)

### `Token` Class
- New class for managing any token-level abstractions
- Added functions for fetching deployment addresses and balances

### Wallet-level Balance Fetching
- implemented `getBalance()` on wallet to get total token balance across all supported chains

### DIFF
--- a/packages/sdk/src/constants/supportedChains.ts
+++ b/packages/sdk/src/constants/supportedChains.ts
@@ -1,0 +1,5 @@
+import { mainnet, unichain } from 'viem/chains'
+
+export const SUPPORTED_CHAIN_IDS = [mainnet.id, unichain.id] as const
+
+export type SupportedChainId = (typeof SUPPORTED_CHAIN_IDS)[number]
--- a/packages/sdk/src/lend/providers/morpho/api.test.ts
+++ b/packages/sdk/src/lend/providers/morpho/api.test.ts
@@ -71,15 +71,17 @@ describe('Morpho API Integration', () => {
 
   it('should handle API errors gracefully', async () => {
     // Mock fetch to simulate network error
-    const originalFetch = global.fetch
-    global.fetch = vi.fn().mockRejectedValue(new Error('Network error'))
+    vi.stubGlobal(
+      'fetch',
+      vi.fn().mockRejectedValue(new Error('Network error')),
+    )
 
     try {
       const vaultData = await fetchRewards(GAUNTLET_USDC_VAULT)
       expect(vaultData).toBeNull()
     } finally {
       // Restore original fetch
-      global.fetch = originalFetch
+      vi.unstubAllGlobals()
     }
   })
 
--- a/packages/sdk/src/services/ChainManager.ts
+++ b/packages/sdk/src/services/ChainManager.ts
@@ -1,0 +1,69 @@
+import { chainById } from '@eth-optimism/viem/chains'
+import { createPublicClient, http, type PublicClient } from 'viem'
+
+import type { SUPPORTED_CHAIN_IDS } from '@/constants/supportedChains.js'
+import type { ChainConfig } from '@/types/chain.js'
+
+/**
+ * Chain Manager Service
+ * @description Manages public clients and chain infrastructure for the Verbs SDK
+ */
+export class ChainManager {
+  private publicClients: Map<(typeof SUPPORTED_CHAIN_IDS)[number], PublicClient>
+  private chainConfigs: ChainConfig[]
+
+  constructor(chains: ChainConfig[]) {
+    this.chainConfigs = chains
+    this.publicClients = this.createPublicClients(chains)
+  }
+
+  /**
+   * Create public clients for all configured chains
+   */
+  private createPublicClients(
+    chains: ChainConfig[],
+  ): Map<(typeof SUPPORTED_CHAIN_IDS)[number], PublicClient> {
+    const clients = new Map<
+      (typeof SUPPORTED_CHAIN_IDS)[number],
+      PublicClient
+    >()
+
+    for (const chainConfig of chains) {
+      const chain = chainById[chainConfig.chainId]
+      if (!chain) {
+        throw new Error(`Chain not found for ID: ${chainConfig.chainId}`)
+      }
+      if (clients.has(chainConfig.chainId)) {
+        throw new Error(
+          `Public client already configured for chain ID: ${chainConfig.chainId}`,
+        )
+      }
+      const client = createPublicClient({
+        chain,
+        transport: http(chainConfig.rpcUrl),
+      })
+
+      clients.set(chainConfig.chainId, client)
+    }
+
+    return clients
+  }
+
+  /**
+   * Get public client for a specific chain
+   */
+  getPublicClient(chainId: (typeof SUPPORTED_CHAIN_IDS)[number]): PublicClient {
+    const client = this.publicClients.get(chainId)
+    if (!client) {
+      throw new Error(`No public client configured for chain ID: ${chainId}`)
+    }
+    return client
+  }
+
+  /**
+   * Get supported chain IDs
+   */
+  getSupportedChains() {
+    return this.chainConfigs.map((c) => c.chainId)
+  }
+}
--- a/packages/sdk/src/types/wallet.ts
+++ b/packages/sdk/src/types/wallet.ts
@@ -1,6 +1,11 @@
 import type { Address } from 'viem'
 
-import type { LendOptions, LendProvider, LendTransaction } from './lend.js'
+import type {
+  LendOptions,
+  LendProvider,
+  LendTransaction,
+} from '@/types/lend.js'
+import type { TokenBalance } from '@/types/token.js'
 
 /**
  * Wallet provider interface
@@ -61,10 +66,10 @@ export interface GetAllWalletsOptions {
  */
 export type WalletVerbs = {
   /**
-   * Get wallet balance
-   * @returns Promise resolving to wallet balance in wei
+   * Get asset balances aggregated across all supported chains
+   * @returns Promise resolving to array of asset balances
    */
-  getBalance(): Promise<bigint>
+  getBalance(): Promise<TokenBalance[]>
   /**
    * Lend assets to a lending market
    * @param asset - Asset token address to lend
--- a/packages/sdk/src/verbs.ts
+++ b/packages/sdk/src/verbs.ts
@@ -1,28 +1,16 @@
 import { createPublicClient, http, type PublicClient } from 'viem'
-import { mainnet, optimism } from 'viem/chains'
+import { mainnet, optimism, unichain } from 'viem/chains'
 
-import { LendProviderMorpho } from './lend/index.js'
-import type { LendProvider } from './types/lend.js'
-import type { VerbsConfig, VerbsInterface } from './types/verbs.js'
-import type { GetAllWalletsOptions, WalletProvider } from './types/wallet.js'
-import type { Wallet } from './wallet/index.js'
-import { WalletProviderPrivy } from './wallet/providers/privy.js'
-
-// Unichain configuration
-const unichain = {
-  id: 130,
-  name: 'Unichain',
-  nativeCurrency: { name: 'Ether', symbol: 'ETH', decimals: 18 },
-  rpcUrls: {
-    default: { http: ['https://mainnet.unichain.org/'] },
-  },
-  blockExplorers: {
-    default: {
-      name: 'Unichain Explorer',
-      url: 'https://unichain.blockscout.com',
-    },
-  },
-}
+import { LendProviderMorpho } from '@/lend/index.js'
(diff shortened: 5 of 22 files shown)


tremarkley:
* **#30** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/30?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#28** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/28?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/28?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/sdk/src/constants/tokenAddresses.ts
```diff
@@ -0,0 +1,9 @@
+import type { Address, Chain } from 'viem'
+import { unichain } from 'viem/chains'
+
+/**
+ * USDC token addresses by chain ID
+ */
+export const USDC_TOKEN_ADDRESSES: Record<number, Address> = {
+  [unichain.id]: '0x078D782b760474a361dDA0AF3839290b0EF57AD6',
+}
```
its-everdred: 👍 this organization makes sense. We don't need to do it in this PR, but I can see use eventually surfacing a `SUPPORTED_TOKENS` variable that can be configured by the sdk consumer.

packages/sdk/src/constants/supportedChains.ts
```diff
@@ -0,0 +1,4 @@
+import type { Chain } from 'viem'
+import { unichain } from 'viem/chains'
+
+export const SUPPORTED_CHAINS: Chain[] = [unichain] as const
```
its-everdred: 👍 great for the first pass, easy to expand later

packages/sdk/src/services/chainManager.ts
```diff
@@ -0,0 +1,77 @@
...
+    const clients = new Map<number, PublicClient>()
+
+    for (const chainConfig of chains) {
+      const client = createPublicClient({
+        chain: SUPPORTED_CHAINS.find(chain => chain.id === chainConfig.chainId),
+        transport: http(chainConfig.rpcUrl),
+      })
+
+      clients.set(chainConfig.chainId, client)
+    }
```
its-everdred: 🤔 I think we should throw if an unsupported chain is passed in

packages/sdk/src/types/verbs.ts
```diff
@@ -32,6 +33,21 @@ export interface VerbsInterface {
...
+ * Chain configuration
+ * @description Configuration for each supported chain
+ */
+export interface ChainConfig {
+  /** Chain ID */
+  chainId: (typeof SUPPORTED_CHAINS)[number]['id']
+  /** RPC URL for the chain */
+  rpcUrl: string
+  /** Optional chain name for display */
+  name?: string
```
its-everdred: 🤔 I think viem stores something like a `name` field. I wouldn't be opposed to making this required and tying it to that value

packages/sdk/src/types/wallet.ts
```diff
@@ -47,14 +47,38 @@ export interface GetAllWalletsOptions {
...
+  getTotalBalance(): Promise<bigint>
+
+  /**
+   * Get USDC balance for a specific chain
+   * @param chainId - Target chain ID
+   * @returns Promise resolving to balance in wei
    */
-  getBalance(): Promise<bigint>
+  getBalanceForChain(chainId: number): Promise<bigint>
 }
```
its-everdred: 🤔 Even though we'll only support USDC at first, I'd like us to avoid baking this assumption in too deep from the start. Up to you if you think it makes sense to address now or later.

packages/sdk/src/wallet/index.ts
```diff
@@ -9,22 +15,76 @@ import type { Wallet as WalletInterface } from '../types/wallet.js'
...
+        console.warn(`Failed to fetch balance for chain ${chainId}:`, error)
+        return {
+          chainId,
+          balance: 0n,
+        }
+      }
+    })
+
+    return Promise.all(balancePromises)
   }
```
its-everdred, line 38: 🤔 I keep going back and forth in my head about what the right level of abstraction is for chainId. I want verbs to mostly obscure chains from the caller. the SDK consumer should be able to opt into chain complexity if they want, but the default should feel hidden and interoperable.

With that in mind, this return type feels surprising for this function. A web2 dev would expect wallet.getBalance() to return something like a mapping of assets to amounts, maybe eventually with prices (we don't need to implement that just yet). 

But I do suggest either renaming this `getBalanceByChains` and keeping it private and/or, updating this to return a summed total balance across all changes.

its-everdred (commented):
Left some early thoughts. The other unknown for this PR is how the frontend -> backend will identify the target wallet. One privy sdk function I saw required a privy walletID, which I'm not sure if we want the backend to have to store that.

Ideally, the backend can provide a unique ID, like an email or user ID, at creation and then re-use that to fetch the wallet in subsequent calls.

We may decide to throw in some very quick state management in the backend to map that userID to the wallet address, but I don't want to set up a db until we're sure it's necessary, so maybe there's a way to be clever to avoid it

packages/sdk/src/types/verbs.ts
tremarkley: yeah, I think for now we can leave this out, espcially since this is for the sdk and not a ui library

packages/sdk/src/types/wallet.ts
tremarkley: updated to not bake in any assumptions around the assets supported!

packages/sdk/src/constants/tokenAddresses.ts
tremarkley: I ended up updating this and introducing a `Token` class and a `TokenRegistry` service

packages/sdk/src/services/chainManager.ts
tremarkley: updated!

packages/sdk/src/wallet/index.ts
tremarkley: yeah, I agree with you here. I updated this so that the public `getBalance` function returns the aggregation of the token balance across all chains

packages/sdk/src/wallet/providers/privy.ts
```diff
@@ -55,8 +56,8 @@ export class PrivyWalletProvider implements WalletProvider {
       // TODO: Implement proper user-to-wallet lookup
```
its-everdred, line 68: I think this TODO might still be true and need work either in this PR or the next. If I initialize a wallet with ID `123` and later pass `123` to privy's getWallet function, I _think_ it'll error because they expect their own walletId and not our arbitrary userId.

packages/sdk/src/services/ChainManager.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @description Manages public clients and chain infrastructure for the Verbs SDK
+ */
+export class ChainManager {
+  private publicClients: Map<(typeof SUPPORTED_CHAIN_IDS)[number], PublicClient>
+  private chainConfigs: ChainConfig[]
+
+  constructor(chains: ChainConfig[]) {
+    this.chainConfigs = chains
+    this.publicClients = this.createPublicClients(chains)
+  }
```
its-everdred, line 18: 🙌 This class will already be useful in my branch

packages/sdk/src/services/TokenRegistry.ts
```diff
@@ -0,0 +1,22 @@
...
+   * Map of token symbols to token instances
+   */
+  private tokens = new Map<string, Token>()
+
+  /**
+   * Register a token instance
+   */
+  register(token: Token): void {
+    this.tokens.set(token.symbol, token)
+  }
```
its-everdred: 🤔 I'm not sure I love registering + storing all supported tokens in memory vs just keeping a static config. I'm down to chat through this

packages/sdk/src/tokens/USDC.ts
```diff
@@ -0,0 +1,18 @@
...
+  constructor(chainManager: ChainManager) {
+    super(chainManager, 'USDC')
+  }
+
+  protected chainAddresses(): Record<SupportedChainId, Address | undefined> {
+    return {
+      [unichain.id]: '0x078D782b760474a361dDA0AF3839290b0EF57AD6',
+    }
+  }
+}
```
its-everdred: 🤔 Continuation of this [comment](https://github.com/ethereum-optimism/verbs/pull/28/files#r2240421540), having to create classes for every supported token feels heavy. I could see replacing this with a static const SUPPORTED_TOKENS and then we don't have to worry about registering and loading them all into memory

its-everdred: In general, I think we're still feeling around to determine _how_ object oriented verbs should be, and which objects should be stateful vs static

packages/sdk/src/verbs.ts
```diff
@@ -32,11 +50,24 @@ export class Verbs implements VerbsInterface {
 
     switch (wallet.type) {
       case 'privy':
-        return new PrivyWalletProvider(wallet.appId, wallet.appSecret)
+        return new PrivyWalletProvider(
+          wallet.appId,
+          wallet.appSecret,
+          this.tokenRegistry,
```
its-everdred: A static token config also saves us from needing to pass this in

packages/sdk/src/wallet/providers/privy.ts
```diff
@@ -55,8 +56,8 @@ export class PrivyWalletProvider implements WalletProvider {
...
       const wallet = await this.privy.walletApi.getWallet({ id: userId })
 
-      const walletInstance = new Wallet(wallet.id)
-      walletInstance.address = wallet.address as Address
+      const walletInstance = new Wallet(wallet.id, this.tokenRegistry)
+      walletInstance.init(getAddress(wallet.address))
```
its-everdred: I could see us eventually rewriting the wallet creation code, it's a bit intuitive.

packages/sdk/src/wallet/providers/privy.ts
```diff
@@ -77,8 +78,8 @@ export class PrivyWalletProvider implements WalletProvider {
       })
 
       return response.data.map((wallet) => {
-        const walletInstance = new Wallet(wallet.id)
-        walletInstance.address = wallet.address as Address
+        const walletInstance = new Wallet(wallet.id, this.tokenRegistry)
```
its-everdred: A static token config also saves us from needing to pass this in here as well

packages/sdk/src/services/TokenRegistry.ts
its-everdred: To be more specific, I could see replacing this class with something like:
```
export const SUPPORTED_TOKENS: Record<string, TokenInfo> = {
  USDC: {
    symbol: 'USDC',
    name: 'USDC',
    decimals: 6,
    addresses: {
      1: '0xA0b86a33E6416eFB1e57D696bDc080e07a4aE3d1', // Ethereum
      130: '0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913', // Unichain
    },
  }
}   
```
or an array instead of a hash

packages/sdk/src/wallet/providers/privy.ts
tremarkley: ok sounds good, will look into this in a separate PR since this one is already getting pretty big

packages/sdk/src/tokens/USDC.ts
tremarkley: yeah that's fair. I'm down to have tokens be statically defined for now and wait till we have a better idea of what abstraction makes the most sense

packages/sdk/src/lend/providers/morpho/api.test.ts
```diff
@@ -71,15 +71,17 @@ describe('Morpho API Integration', () => {
...
+      vi.fn().mockRejectedValue(new Error('Network error')),
+    )
 
     try {
       const vaultData = await fetchRewards(GAUNTLET_USDC_VAULT)
       expect(vaultData).toBeNull()
     } finally {
       // Restore original fetch
-      global.fetch = originalFetch
+      vi.unstubAllGlobals()
```
its-everdred, line 84: nice!

its-everdred (approved):

PR #32, opened 2025-07-30
Wallet lending

its-everdred:
## Problem

We need a lend deposit function

## Solution

- [x] Initialize the `ChainManager` and `LendProvider` in `Verbs` class instead of `Wallet`
- [x] Add supersim to integration tests, fork Unichain and test 
- [x] Implement wallet.lend()
- [x] Test on unichain
- [x] /lend/deposit
- [x] /wallet/send
- [x] /lend/vault/123/balance/abc
- [x] Start clean up

Finish clean up moved to this [ticket](https://github.com/ethereum-optimism/verbs/issues/36)


.circleci/config.yml
```diff
@@ -82,6 +82,42 @@ jobs:
           name: Run Backend Tests
           command: cd packages/demo/backend && pnpm test
 
+  supersim-tests:
```
tremarkley: I would recommending using mise to install foundry and supersim as dependencies. We use mise throughout the org and here is an example of how it is configured for the ecosystem repo: https://github.com/ethereum-optimism/ecosystem/blob/main/mise.toml and https://github.com/ethereum-optimism/ecosystem/blob/ae1a6343a969a65c2f0880c201631bc8a8d226ce/.circleci/config.yml#L25-L48

packages/sdk/src/lend/providers/morpho/lend.supersim.test.ts
```diff
@@ -0,0 +1,278 @@
...
+      },
+      chainId: unichain.id,
+      rpcUrl: 'http://127.0.0.1:9546',
+    })
+
+    // For testing, create a wallet directly with the Verbs instance
+    // In real app, wallet.lend() would be available after createWallet()
+    const { Wallet } = await import('../../../wallet/index.js')
+    testWallet = new Wallet('test-wallet', verbs)
+    testWallet.init(testAccount.address)
```
tremarkley: this api feels a little unintuitive / heavy to to me,. what do you think of having a `createWallet` function under the verbs instance, seems like we could simplify this and do `verbs.createWallet()`? Also, why is the import here instead of at the top of the file?

packages/sdk/src/lend/providers/morpho/lend.supersim.test.ts
```diff
@@ -0,0 +1,278 @@
...
+    // Check that we can connect and get the chain ID
+    const chainId = await publicClient.getChainId()
+    expect(chainId).toBe(130) // Unichain chain ID
+
+    // Check that our test wallet has ETH
+    const balance = await publicClient.getBalance({
+      address: testAccount.address,
+    })
+    expect(balance).toBeGreaterThan(0n)
+    console.log(`Test wallet balance: ${balance / 10n ** 18n} ETH`)
```
tremarkley: nit: maybe youre already planning to do this since this is WIP, but we should avoid console logging this and just check that it is the expected value. same comment goes for all the other console logs in this test

packages/sdk/src/lend/providers/morpho/lend.supersim.test.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  it('should execute lend operation with real Morpho transactions', async () => {
+    // First, verify the vault exists
+    console.log(
+      `Testing with vault: ${TEST_VAULT.name} (${TEST_VAULT_ADDRESS})`,
+    )
+    const vaultInfo = await verbs.lend.getVault(TEST_VAULT_ADDRESS)
+    console.log(`Vault info: ${vaultInfo.name} - APY: ${vaultInfo.apy}%`)
+
+    // Check USDC balance (wrapped in try-catch as USDC might not be at expected address on fork)
```
tremarkley: why would the USDC not be there on the fork?

packages/sdk/src/lend/providers/morpho/lend.supersim.test.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    console.log('✅ All transaction data validation passed!')
+
+    // Test actual transaction sending with approval call data
+    console.log('Testing approval transaction...')
+
+    try {
+      // Note: This will likely fail on forked networks without actual USDC,
+      // but we can test the transaction structure and gas estimation
+      const approvalTx = lendTx.transactionData!.approval!
```
tremarkley: if you want to give the test account USDC or make changes to balances, I believe you should be able to do that through the `testClient`, here is how we set up test clients in the ecosystem repo for our supersim tests: https://github.com/ethereum-optimism/ecosystem/blob/main/packages/viem/src/test/clients.ts. I think you probably want to fork at a specific height so that the balances and deployment addresses dont change: https://github.com/ethereum-optimism/supersim/blob/955c47b34089126f2622e964d79a3cbf1d19ed8f/docs/src/reference/supersim-fork.md?plain=1#L55

packages/sdk/src/verbs.ts
```diff
@@ -82,8 +90,7 @@ export class Verbs implements VerbsInterface {
         return new WalletProviderPrivy(
           wallet.appId,
           wallet.appSecret,
-          this.chainManager,
-          this.lendProvider,
+          this, // Pass Verbs instance so wallets can access configured providers
```
tremarkley, line 98: hmm, I think we should just explicitly pass in any dependencies we need. reason being is that for testing purposes it's much easier to just mock the dependencies you need and pass them in. wdyt?

packages/sdk/src/types/lend.ts
```diff
@@ -2,6 +2,19 @@ import type { Address } from 'viem'
...
 
+/**
+ * Transaction data for execution
+ * @description Raw transaction data for wallet execution
+ */
+export interface TransactionData {
+  /** Target contract address */
+  to: Address
+  /** Encoded function call data */
+  data: string
```
tremarkley: nit: should use the `Hex` type here

packages/sdk/src/lend/providers/morpho/lend.supersim.test.ts
```diff
@@ -0,0 +1,278 @@
...
+        type: 'privy',
+        appId: 'test-app-id',
+        appSecret: 'test-app-secret',
+      },
+      lend: {
+        type: 'morpho',
+        defaultSlippage: 50,
+      },
+      chainId: unichain.id,
+      rpcUrl: 'http://127.0.0.1:9546',
```
tremarkley: should we remove these as arguments to verbs and replace it with a `chainsConfig?` that can plug into where we create the `ChainManager`: https://github.com/ethereum-optimism/verbs/blob/ca59c1b16e30c01c0166571c73bc888daef6ea16/packages/sdk/src/verbs.ts#L30?

packages/sdk/src/utils/assets.ts
```diff
@@ -0,0 +1,141 @@
...
+ * @param chainId - Chain ID to resolve for
+ * @returns Resolved asset information
+ * @throws Error if asset is not supported or found
+ */
+export function resolveAsset(
+  asset: AssetIdentifier,
+  chainId: SupportedChainId,
+): ResolvedAsset {
+  // If it's an address (starts with 0x), validate and find symbol
+  if (asset.startsWith('0x')) {
```
tremarkley, line 33: can use the `isAddress` function from viem for this

packages/sdk/src/utils/assets.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/**
+ * Resolve asset identifier to address and metadata
+ * @param asset - Asset symbol (e.g. 'usdc') or address
+ * @param chainId - Chain ID to resolve for
+ * @returns Resolved asset information
+ * @throws Error if asset is not supported or found
+ */
+export function resolveAsset(
```
tremarkley, line 28: I do think the `TokenRegistry` + `Token` class abstraction helps simplify and make this code a bit more maintainable, but I also get if we dont want the extra overhead of classes. Definitely not a requirement to refactor now but we should keep an eye on this and see how it progresses. If this file starts to get pretty big or overly complex we should consider bringing those abstractions back.

packages/sdk/src/utils/assets.ts
```diff
@@ -0,0 +1,141 @@
...
+    throw new Error(
+      `Asset ${asset} is not supported on chain ${chainId}. Available chains: ${Object.keys(tokenInfo.addresses).join(', ')}`,
+    )
+  }
+
+  return {
+    address,
+    symbol: tokenInfo.symbol,
+    decimals: tokenInfo.decimals,
+  }
```
tremarkley, line 79: nit: could break this out into a separate function `resolveBySymbol` and then above could be broken out into `resolveByAsset`

tremarkley (commented):
I like the direction this is going in, just leaving some initial thoughts 👌

.circleci/config.yml
its-everdred: done

packages/demo/backend/src/controllers/wallet.ts
```diff
@@ -161,11 +161,30 @@ export class WalletController {
       if (validation.error) return validation.error
 
       const { userId } = validation.data
-
-      // Parse request body for token type
       const body = await c.req.json().catch(() => ({}))
       const tokenType = body.tokenType || 'USDC'
 
+      // Check if we're running against supersim locally
```
its-everdred:
```suggestion
      // TODO: Find a better way to do this
```

packages/sdk/src/utils/assets.ts
its-everdred: I don't like this file at all and have added a step to delete it in the clean up [ticket](https://github.com/ethereum-optimism/verbs/issues/36)

packages/sdk/src/verbs.ts
its-everdred: I added a step to refactor our verbs sdk interfaces, including this, in our clean up [ticket](https://github.com/ethereum-optimism/verbs/issues/36)

PR #96, opened 2025-09-08
feat: refactor wallet fetching and creation

### LINKED ISSUE #87: Wallet adapter for Privy

### LINKED ISSUE #88: Remove the create wallet functionality for Privy from the Verbs sdk
Hosted wallet creation will always occur outside of the Verbs sdk

### LINKED ISSUE #89: `Wallet` abstract class

tremarkley:
Closes https://github.com/ethereum-optimism/verbs/issues/87
Closes https://github.com/ethereum-optimism/verbs/issues/88
Closes https://github.com/ethereum-optimism/verbs/issues/89

## Changes
- Introduce `HostedWalletProvider.toVerbsWallet(params) → Promise<Wallet>` as the single hosted to Verbs wallet entrypoint
- Drop create/get/getAll privy wallet methods from `PrivyHostedWalletProvider` and delegate this responsibility to the consumer of the Verbs SDK
- Add `toVerbsWallet` to `PrivyHostedWalletProvider` so that Privy wallets can be converted to verb's `Wallet`'s
- Create `Wallet` class: this will be the base wallet class that all Verbs wallets extend from
- Updated `DefaultSmartWallet` class to use a factory pattern for creating instances of `DefaultSmartWallet`, so that async actions can be completed prior to initialization.
- Updated `PrivyWallet` class to use a factory pattern for creating instances of `PrivyWallet`, so that async actions can be completed prior to initialization.
- Updated `PrivyWallet` to extend from `Wallet` class
- Updated `SmartWallet` to extend from `Wallet` class


tremarkley:
* **#99** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/99?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#96** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/96?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/96?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/demo/backend/src/mocks/MockPrivyClient.ts
```diff
@@ -0,0 +1,118 @@
...
+ * @param appId - Mock app ID
+ * @param appSecret - Mock app secret
+ * @returns MockPrivyClient cast as PrivyClient
+ */
+export function createMockPrivyClient(
+  appId: string,
+  appSecret: string,
+): PrivyClient {
+  return new MockPrivyClient(appId, appSecret) as unknown as PrivyClient
+}
```
its-everdred, line 118: 👍  Good call creating a Mock file for this

packages/demo/backend/src/services/lend.ts
```diff
@@ -61,14 +61,13 @@ export async function getVaultBalance(
...
-  const { wallet } = await getWallet(walletId)
+  const wallet = await getWallet(walletId)
 
   if (!wallet) {
     throw new Error(`Wallet not found for user ID: ${walletId}`)
   }
 
-  const walletAddress = await wallet.getAddress()
-  return await verbs.lend.getVaultBalance(vaultAddress, walletAddress)
+  return verbs.lend.getVaultBalance(vaultAddress, wallet.address)
```
its-everdred, line 70: 👍 note: my work will refactor this to `wallet.lend.balance(vaultAddress)`

packages/demo/backend/src/services/wallet.ts
```diff
@@ -13,64 +10,93 @@ import {
...
 export async function getAllWallets(
   options?: GetAllWalletsOptions,
 ): Promise<Array<{ wallet: SmartWallet; id: string }>> {
   try {
     const verbs = getVerbs()
-    const privyWallets = await (
-      verbs.wallet.hostedWalletProvider as PrivyHostedWalletProvider
-    ).getAllWallets(options)
+    const privyClient = getPrivyClient()
+    const response = await privyClient.walletApi.getWallets(options)
```
its-everdred, line 86: 👍 nice, verbs should not define a `getAllWallets()` function

packages/demo/backend/src/services/wallet.ts
```diff
@@ -13,64 +10,93 @@ import {
...
-          signer: account,
-          ownerIndex: 0,
+      response.data.map(async (privyWallet) => {
+        const verbsPrivyWallet = await verbs.wallet.hostedWalletToVerbsWallet({
+          walletId: privyWallet.id,
+          address: privyWallet.address,
+        })
+        const wallet = await verbs.wallet.getSmartWallet({
+          signer: verbsPrivyWallet.signer,
+          deploymentOwners: [getAddress(privyWallet.address)],
```
its-everdred, line 95: 🤔 Instead of the dev calling:
```
const privyWallet = verbs.wallet.hostedWalletToVerbsWallet(...)
const wallet = await verbs.wallet.getSmartWallet({...})
```
could `hostedWalletToVerbsWallet` return the smartWallet as well? like
```
const privyWallet = verbs.wallet.hostedWalletToVerbsWallet(...)
const wallet = privyWallet.smartWallet()
```

Separate from this specific case, but a general thought I've had, I'm worried about devs juggling context between or even confusing HostedWallets, PrivyWallets, VerbsWallets, SmartWallets. Our goal should be to make it *very* difficult for a dev to accidentally put money in the user's wrong sub-wallet, though I'm not sure how to achieve that goal.

packages/sdk/src/wallet/WalletNamespace.ts
```diff
@@ -67,52 +56,17 @@ export class WalletNamespace {
...
+   * @param params.walletId - Unique identifier for the hosted wallet
+   * @param params.address - Ethereum address of the hosted wallet
+   * @returns Promise resolving to the Verbs wallet instance
    */
-  async getHostedWallet(params: getHostedWalletOptions) {
-    return this.provider.getHostedWallet(params)
+  async hostedWalletToVerbsWallet(
+    params: HostedWalletToVerbsWalletOptions,
+  ): Promise<VerbsWallet> {
+    return this.provider.hostedWalletToVerbsWallet(params)
```
its-everdred, line 69: 🤔 Not necessary for this PR, but I wonder if there's a way to abstract this file away since it's just calls the provider versions of all functions.

packages/sdk/src/wallet/base/VerbsWallet.ts
```diff
@@ -0,0 +1,81 @@
+import type { Address, LocalAccount, WalletClient } from 'viem'
+
+import type { SupportedChainId } from '@/constants/supportedChains.js'
+
+/**
+ * Base verbs wallet class
+ * @description Abstract base class for verbs wallet implementations.
+ * Provides a standard interface for verbs wallets.
+ */
+export abstract class VerbsWallet {
```
its-everdred: 🤔 What is the role of this class? Continuing my thought from the other comment, I wondering if surfacing this interface to devs via `toVerbsWallet` functions will lead to confusion? If we intend to introduce this class, we should document/differentiate alongside the others in the architecture doc.

packages/demo/backend/src/services/wallet.ts
tremarkley: I'm open to adding that if there is a preference from users to do that and this ends up being a painpoint that is easy for them to get wrong. 

However, I actually think that converting a hostedWallet automatically to a smart wallet is more likely to result in money being accidentally put in the wrong sub-wallet. By separating this into two separate functions and not trying to combine so many side effects into one action, it forces the dev to be explicit about the configuration of their smart wallet. Also, if we were to combine both into one function we'd have to overload the `hostedWalletToVerbsWallet` with the arguments that `getSmartWallet` uses and the interface becomes a bit confusing.

packages/sdk/src/wallet/WalletNamespace.ts
tremarkley: yeah, I was thinking long term this architecture makes sense if there are things we want to add to `WalletNamespace` that are outside of the `WalletProvider`, but this could definitely be overengineered

packages/sdk/src/wallet/base/VerbsWallet.ts
tremarkley: this is meant to be the class that every wallet extends from. The thinking is that this is the base wallet and then things like the smart wallet, privy wallet, etc all have implementations of this and can add additional functionality on top of it (e.g. [SmartWallet](https://github.com/ethereum-optimism/verbs/blob/4960eb4cd21a1b466403105073c3f5d5bf6e972e/packages/sdk/src/wallet/base/SmartWallet.ts), [PrivyWallet](https://github.com/ethereum-optimism/verbs/blob/4960eb4cd21a1b466403105073c3f5d5bf6e972e/packages/sdk/src/wallet/PrivyWallet.ts))

tremarkley: like we just talked about I am going to rename this to just `Wallet`

packages/demo/backend/src/services/wallet.ts
its-everdred: make sense to me, lets keep it

its-everdred (approved):

PR #107, opened 2025-09-11
feat: type-safe hosted wallet generics

tremarkley:
## Description
This PR introduces first-class, compile-time typing of the hosted wallet provider across the SDK

## Changes
- Make `VerbsConfig` and `HostedWalletConfig` generic over `HostedProviderType` and replace unions with a typed `ProviderSpec`
- Add `HostedProviderInstanceMap` and update `HostedProviderFactory` to return concrete provider instances keyed by type.
- `Generic`-ize `WalletProvider` and `WalletNamespace` to propagate concrete hosted/smart provider types.

## Why
- Ensures end-to-end type safety for hosted providers (no `as` casts).
- Improves pluggability: adding a new provider updates a central type map and preserves strong typing.

### DIFF
--- a/packages/demo/backend/src/config/verbs.ts
+++ b/packages/demo/backend/src/config/verbs.ts
@@ -4,9 +4,9 @@ import { baseSepolia, unichain } from 'viem/chains'
 
 import { env } from './env.js'
 
-let verbsInstance: Verbs
+let verbsInstance: Verbs<'privy'>
 
-export function createVerbsConfig(): VerbsConfig {
+export function createVerbsConfig(): VerbsConfig<'privy'> {
   return {
     wallet: {
       hostedWalletConfig: {
@@ -53,7 +53,7 @@ export function createVerbsConfig(): VerbsConfig {
   }
 }
 
-export function initializeVerbs(config?: VerbsConfig): void {
+export function initializeVerbs(config?: VerbsConfig<'privy'>): void {
   const verbsConfig = config || createVerbsConfig()
   verbsInstance = new Verbs(verbsConfig)
 }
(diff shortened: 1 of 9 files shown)


tremarkley:
* **#107** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/107?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/demo/backend/src/config/verbs.ts
```diff
@@ -53,7 +53,7 @@ export function createVerbsConfig(): VerbsConfig {
   }
 }
 
-export function initializeVerbs(config?: VerbsConfig): void {
+export function initializeVerbs(config?: VerbsConfig<'privy'>): void {
```
its-everdred, line 56: ❓ what does this type eventually look like when a dev has to select providers for wallet, lend, swap etc?

its-everdred (approved):
🙌

packages/demo/backend/src/config/verbs.ts
tremarkley: there would end up being an entry for each one so it would look like `VerbsConfig<'privy', 'morpho',...etc>`

we can find ways to improve that devx, but also I wouldnt expect devs to actually need to fill in the generics themselves, they should be able to just specify their config as a const and then ts will be able to infer all the generics for them. This is only needed because the demo app doesnt specify the config inline.

PR #115, opened 2025-09-15
rename vault -> market

its-everdred:
## Problem 

We use both "vault" and "market". 

## Solution

Consolidate all instances to "market"


packages/demo/backend/src/controllers/lend.ts
```diff
@@ -16,15 +16,7 @@ const DepositRequestSchema = z.object({
...
 
-const VaultAddressParamSchema = z.object({
-  params: z.object({
-    vaultAddress: z
-      .string()
-      .regex(/^0x[a-fA-F0-9]{40}$/, 'Invalid vault address format'),
-  }),
-})
-
-const VaultBalanceParamsSchema = z.object({
```
its-everdred, line 29: This was unused

tremarkley (approved):

PR #119, opened 2025-09-17
Refactor Asset + Market interfaces

its-everdred:
## Problem

We need to allow users to define their own Assets and Markets

## Solution

- [x] Move Asset definitions sdk -> backend
- [x] Move LendMarketConfig definitions sdk -> backend
- [x] Make verbs.lend.config public
- [x] Refactor and increase test converage


packages/demo/backend/src/config/markets.ts
```diff
@@ -0,0 +1,32 @@
...
+export const USDCDemoVault: LendMarketConfig = {
+  address: '0x297E324C46309E93112610ebf35559685b4E3547' as Address,
+  chainId: baseSepolia.id,
+  name: 'USDC Demo Vault (Base Sepolia)',
+  asset: USDC_DEMO,
+  lendProvider: 'morpho',
+}
+
+// Export all markets for easy consumption
+export const ALL_MARKETS = [GauntletUSDC, MetaMorphoUSDC, USDCDemoVault]
```
its-everdred, line 32: ⚠️ I need to double check these values

packages/demo/backend/src/config/assets.ts
```diff
@@ -0,0 +1,56 @@
...
+    [base.id]: 'native',
+    [baseSepolia.id]: 'native',
+  },
+  metadata: {
+    decimals: 18,
+    name: 'Ethereum',
+    symbol: 'ETH',
+  },
+  type: 'native',
+}
```
its-everdred, line 17: ⚠️ ETH and WETH are likely to cause problems and will require a lot of edge case handling. We may also move this into an enshrined Asset definition in the sdk.

packages/demo/backend/src/config/verbs.ts
```diff
@@ -29,7 +29,8 @@ export function createVerbsConfig(): VerbsConfig<'privy'> {
       },
     },
     lend: {
-      type: 'morpho',
+      provider: 'morpho',
```
its-everdred, line 32: I like `provider` over `type` in the config

packages/sdk/src/lend/namespaces/VerbsLendNamespace.ts
```diff
@@ -10,6 +10,14 @@ import type { LendMarket, LendMarketId } from '@/types/lend.js'
 export class VerbsLendNamespace {
   constructor(protected readonly provider: LendProvider) {}
 
+  /**
+   * Get lending provider configuration
+   * @description Access to provider configuration including defaultSlippage, provider type, etc.
+   */
+  get config(): LendConfig {
+    return this.provider.config as LendConfig
+  }
```
its-everdred, line 24: 🤔 Thoughts about making this publicly accessible so the dev can reference their ingested config. options:

1. `verbs.lend.config` - config + decorators
2. `verbs.config` - config + decorators
3. skip, the dev can just access their config directly and we wont need to augment it at runtime

packages/sdk/src/lend/providers/morpho/index.test.ts
```diff
@@ -202,4 +202,140 @@ describe('LendProviderMorpho', () => {
...
+            metadata: {
+              decimals: 18,
+              name: 'Wrapped Ether',
+              symbol: 'WETH',
+            },
+            type: 'erc20',
+          },
+          lendProvider: 'morpho',
+        },
+      ]
```
its-everdred, line 322: This can be cleaned up

packages/sdk/src/lend/providers/morpho/vaults.ts
```diff
@@ -144,15 +148,51 @@ export function calculateBaseApy(vault: any): number {
...
   chainManager: ChainManager,
+  marketAllowlist?: LendMarketConfig[],
 ): Promise<LendMarket> {
+  let config: LendMarketConfig | undefined
+  let vaultAddress: Address | undefined
+
   try {
-    // 1. Find vault configuration for validation
-    const config = SUPPORTED_VAULTS.find((c) => c.address === vaultAddress)
+    if (typeof vaultConfig === 'string') {
```
its-everdred, line 159: this is ugly, will fix

packages/sdk/src/lend/providers/morpho/vaults.ts
```diff
@@ -144,15 +148,51 @@ export function calculateBaseApy(vault: any): number {
  * @returns Promise resolving to detailed vault information
  */
 export async function getVaultInfo(
-  vaultAddress: Address,
+  vaultConfig: Address | LendMarketConfig,
```
its-everdred, line 151: will change this to two separate param options and require one: `vaultConfig?, address?`

packages/sdk/src/wallet/base/Wallet.spec.ts
```diff
@@ -60,15 +59,8 @@ describe('Wallet (base)', () => {
...
-    expect(fetchERC20Balance).toHaveBeenCalledTimes(tokenCount)
-
-    // Ensure each call used the same chainManager and address
-    for (const call of (fetchERC20Balance as Mock).mock.calls) {
-      expect(call[0]).toBe(chainManager)
-      expect(call[1]).toBe(address)
-      expect(call[2]).toBeTruthy()
-    }
+    // Currently only ETH balance is fetched
+    expect(fetchERC20Balance).toHaveBeenCalledTimes(0)
```
its-everdred, line 63: revert

packages/sdk/src/wallet/base/Wallet.ts
```diff
@@ -49,14 +49,9 @@ export abstract class Wallet {
...
-    const tokenBalancePromises = Object.values(SUPPORTED_TOKENS).map(
-      async (token) => {
-        return fetchERC20Balance(this.chainManager, this.address, token)
-      },
-    )
+    // For now, just return ETH balance until assets are properly configured
     const ethBalancePromise = fetchETHBalance(this.chainManager, this.address)
-
-    return Promise.all([ethBalancePromise, ...tokenBalancePromises])
+    return Promise.all([ethBalancePromise])
```
its-everdred, line 54: ⚠️ This is not what we want. How do we want to handle the wallet.getBalance() function? Options:
1. Have the dev pass the assets they "support" in VerbsConfig and only check the balances of those
2. Keep a small temporary asset list in the SDK for this function 
3. Tie this function into Optimism's Token List

Leaning toward 1 or 2 for this PR, 3 is pretty involved

packages/sdk/src/wallet/providers/DefaultSmartWalletProvider.ts
```diff
@@ -30,7 +30,7 @@ export class DefaultSmartWalletProvider extends SmartWalletProvider {
    */
   constructor(
     chainManager: ChainManager,
-    lendProvider: LendProvider,
+    lendProvider?: LendProvider,
```
its-everdred, line 33: I'm not sure how the WalletProvider <> LendProvider interface should work in the end

packages/sdk/src/wallet/DefaultSmartWallet.ts
```diff
@@ -161,26 +156,41 @@ export class DefaultSmartWallet extends SmartWallet {
...
+      throw new Error('Amount must be greater than 0')
+    }
+
+    // Check if asset is supported on the chain
+    const tokenAddress = asset.address[chainId]
+    if (!tokenAddress) {
+      throw new Error(
+        `${asset.metadata.symbol} not supported on chain ${chainId}`,
+      )
+    }
```
its-everdred, line 175: This whole function will change later, just wanted to make sure it works with the new Asset shape for now

packages/sdk/src/wallet/DefaultSmartWallet.ts
```diff
@@ -161,26 +156,41 @@ export class DefaultSmartWallet extends SmartWallet {
...
+
+    // Parse human-readable amount
+    const parsedAmount = parseAssetAmount(amount, asset.metadata.decimals)
```
its-everdred, line 179: This whole function will change later, just wanted to make sure it works with the new Asset shape for now

packages/sdk/src/verbs.ts
```diff
@@ -70,15 +68,11 @@ export class Verbs<THostedWalletProviderType extends HostedProviderType> {
...
+   * Get the lend provider instance
+   * @returns LendProvider instance if configured, undefined otherwise
    */
-  get lendProviderInstance(): LendProvider {
-    if (!this.lendProvider) {
-      throw new Error('Lend provider not configured')
-    }
-    return this.lendProvider
+  get lendProvider(): LendProvider | undefined {
+    return this._lendProvider
```
its-everdred, line 81: 🤔 Also not sure how we want this to work in the end, or even if we want it accessible, but this currently makes the LendProvider instance accessible via `verbs.lendProvider`.

packages/demo/backend/src/services/wallet.ts
```diff
@@ -204,5 +205,5 @@ export async function sendTokens(
     throw new Error('Wallet not found')
   }
 
-  return wallet.sendTokens(amount, 'usdc', recipientAddress)
+  return wallet.sendTokens(amount, USDC, baseSepolia.id, recipientAddress)
```
its-everdred, line 212: i dont love hardcoding network here, but this is used by the `wallet send` function, which won't be in the next demo

packages/sdk/src/lend/providers/morpho/sdk.ts
```diff
@@ -137,27 +91,36 @@ export function calculateBaseApy(vault: any): number {
...
  * @returns Promise resolving to detailed vault information
  */
-export async function getVaultInfo(
-  vaultAddress: Address,
-  chainManager: ChainManager,
-): Promise<LendMarket> {
-  try {
-    // 1. Find vault configuration for validation
-    const config = SUPPORTED_VAULTS.find((c) => c.address === vaultAddress)
+export async function getVault(params: GetVaultParams): Promise<LendMarket> {
```
its-everdred, line 110: Going to keep `vault` here since this function lives within `/morpho/sdk.ts` and wraps the morpho vault functions.

packages/sdk/src/constants/assets.ts
```diff
@@ -0,0 +1,18 @@
...
+    [base.id]: '0x0000000000000000000000000000000000000000',
+    [baseSepolia.id]: '0x0000000000000000000000000000000000000000',
+  },
+  metadata: {
+    decimals: 18,
+    name: 'Ethereum',
+    symbol: 'ETH',
+  },
+  type: 'native',
+}
```
its-everdred, line 20: ℹ️ This is important enough to have it live in the SDK from the beginning since we have to have special cases as the native asset. All others can be defines by the backend until we integrate with token list

packages/sdk/src/lend/namespaces/VerbsLendNamespace.ts
tremarkley: I dont have a strong preference on this! nit: is the cast necessary?

packages/sdk/src/lend/providers/morpho/index.ts
```diff
@@ -123,7 +119,10 @@ export class LendProviderMorpho extends LendProvider {
             value: 0n,
           },
         },
-        slippage: options?.slippage || this.defaultSlippage,
+        slippage:
+          options?.slippage ||
+          this._config.defaultSlippage ||
```
tremarkley: nit: you could have `this._config.defaultSlippage` always be defined and default to `DEFAULT_VERBS_CONFIG.lend.defaultSlippage` that way it always has a default value when it isnt defined

packages/sdk/src/services/tokenBalance.ts
```diff
@@ -77,45 +76,51 @@ export async function fetchERC20Balance(
...
 
   // Handle native ETH balance
-  if (token.symbol === 'ETH') {
+  if (asset.type === 'native') {
     return {
       balance: await publicClient.getBalance({
         address: walletAddress,
       }),
-      tokenAddress,
+      tokenAddress: '0x0000000000000000000000000000000000000000' as Address,
```
tremarkley: nit: isnt this address already defined in the `ETH` token? or is there a reason we have to hardcode this?

packages/sdk/src/services/tokenBalance.ts
```diff
@@ -77,45 +76,51 @@ export async function fetchERC20Balance(
...
   }
 
   // Handle ERC20 token balance
+  if (tokenAddress === 'native') {
+    throw new Error('Unexpected native token address for ERC20 asset')
+  }
+
   const balance = await publicClient.readContract({
-    address: tokenAddress,
+    address: tokenAddress as Address,
```
tremarkley, line 116: nit: it might be cleaner to just type `tokenAddress` as an Address type so we dont have to cast it here and below

packages/sdk/src/types/asset.ts
```diff
@@ -0,0 +1,34 @@
+import type { Address } from 'viem'
+
+import type { SupportedChainId } from '@/constants/supportedChains.js'
+
+/**
+ * Asset type inspired by EIP 7811 but adapted for multi-chain asset definitions
+ */
+export interface Asset {
+  /** Multi-chain address mapping - use 'native' for native tokens like ETH */
+  address: Partial<Record<SupportedChainId, Address | 'native'>>
```
tremarkley: question - why can the value be of type `Address | 'native'`? that seems like it could get messy because everywhere we use an address we'll have to check if it's native. typically native ether is just set to the `0x0000000000000000000000000000000000000000` address
Screenshot 2025-09-17 at 5 13 37 PM

packages/sdk/src/verbs.ts
tremarkley: hmm, yeah I think I did something similar with the wallet providers. now that I think about it I wonder if we should keep these private to start in order to minimize the surface of the external api that has to be maintained

packages/sdk/src/wallet/base/Wallet.ts
```diff
@@ -49,9 +49,20 @@ export abstract class Wallet {
...
+        // Convert TokenInfo to Asset format
+        const asset = {
+          address: tokenInfo.addresses,
+          metadata: {
+            decimals: tokenInfo.decimals,
+            name: tokenInfo.name,
+            symbol: tokenInfo.symbol,
+          },
+          type: tokenInfo.symbol === 'ETH' ? 'native' : 'erc20',
+        } as const
```
tremarkley: a potential improvement is we just update the `SUPPORTED_TOKENS` to be type `Asset[]`, that way we dont have to do this conversion everywhere we want to use them, wdyt?

tremarkley (approved):
looks good, just some suggestions!

packages/demo/backend/src/config/assets.ts
```diff
@@ -0,0 +1,41 @@
...
+    [unichain.id]: '0x078d782b760474a361dda0af3839290b0ef57ad6',
+    [baseSepolia.id]: '0x036CbD53842c5426634e7929541eC2318f3dCF7e',
+  },
+  metadata: {
+    decimals: 6,
+    name: 'USDC',
+    symbol: 'USDC',
+  },
+  type: 'erc20',
+}
```
tremarkley, line 16: question - should the verbs sdk just export a `USDC` asset and the user can use that? It feels like a lot of friction and potential room for error to have a dev need to define a commonly used asset like USDC

tremarkley (commented):
one more question

packages/sdk/src/wallet/base/Wallet.ts
```diff
@@ -49,9 +49,20 @@ export abstract class Wallet {
...
   async getBalance(): Promise<TokenBalance[]> {
+    // TEMPORARY - will use optimism token list eventually
     const tokenBalancePromises = Object.values(SUPPORTED_TOKENS).map(
-      async (token) => {
-        return fetchERC20Balance(this.chainManager, this.address, token)
+      async (tokenInfo) => {
+        // Convert TokenInfo to Asset format
+        const asset = {
+          address: tokenInfo.addresses,
+          metadata: {
```
tremarkley: shouldnt this also use tokens that are passed in through the verbs config? for example user might want to use a token that isnt in the verbs sdk yet and they can just define it themselves and pass it in to verbs, right?

tremarkley (commented):
sorry one more lol

packages/sdk/src/lend/namespaces/VerbsLendNamespace.ts
its-everdred: nope, fixed

packages/sdk/src/verbs.ts
its-everdred: going to merge as is and we can continue to think about it

packages/demo/backend/src/config/assets.ts
its-everdred: yes, we should have the sdk export all major tokens. I added this [ticket](https://github.com/ethereum-optimism/verbs/issues/122) for that work.

packages/sdk/src/wallet/base/Wallet.ts
its-everdred: correct, I'm going to update in this [ticket](https://github.com/ethereum-optimism/verbs/issues/122)

packages/sdk/src/wallet/base/Wallet.ts
its-everdred: made this change, though SUPPORTED_TOKENS is temporary and will be ripped out soon

packages/sdk/src/types/asset.ts
its-everdred: good find, i had it as an option for a moment, but removed it

PR #129, opened 2025-09-19
Validate lend functions by supported + configured networks

its-everdred:
## Problem

[Issue](https://github.com/ethereum-optimism/verbs/issues/75): Network validation is done by the lend provider child  implementation.

## Solution
 - [x] Add network supported + configured validation functions 
 - [x] Refactor `LendProvider` parent to do all validation, delegate to protected children functions


packages/sdk/src/lend/namespaces/WalletLendNamespace.ts
```diff
@@ -29,6 +29,7 @@ export class WalletLendNamespace<
   async lendExecute(
     asset: Address,
     amount: bigint,
+    chainId: number,
```
tremarkley: question - should this leverage the `LendMarketId` type? https://github.com/ethereum-optimism/verbs/blob/2c6ba665b8013eab58557a6b5a0e90d1727743be/packages/sdk/src/types/lend.ts#L14

packages/sdk/src/lend/provider.ts
```diff
@@ -50,63 +42,75 @@ export abstract class LendProvider<
...
+   * @param chainId - Chain ID for the transaction
    * @param marketId - Optional specific market ID
    * @param options - Optional lending configuration
    * @returns Promise resolving to lending transaction details
    */
-  abstract lend(
+  async lend(
     asset: Address,
     amount: bigint,
+    chainId: number,
```
tremarkley: same question as above, would it make sense to update this to use the `LendMarketId` type?

packages/sdk/src/lend/provider.ts
```diff
@@ -50,63 +42,75 @@ export abstract class LendProvider<
    * @returns Array of supported network chain IDs
    */
   supportedNetworkIds(): number[] {
```
tremarkley: nit: it is a bit confusing that in some places we call it chain id and in other places it's network id. should we just consolidate to chain id or is there a difference between the two?

packages/sdk/src/lend/provider.ts
```diff
@@ -115,19 +119,146 @@ export abstract class LendProvider<
...
+   * @param chainId - Chain ID to validate
+   * @throws Error if network is not supported
+   */
+  protected validateChainIdSupported(chainId: number): void {
+    if (!this.isNetworkSupported(chainId)) {
+      throw new Error(
+        `Network ${chainId} is not supported. Supported networks: ${this.SUPPORTED_NETWORK_IDS.join(', ')}`,
+      )
+    }
+  }
```
tremarkley: similar question here, should we consolidate these? I think having two functions that both do the same thing has potential to create confusion

packages/sdk/src/lend/provider.ts
```diff
@@ -115,19 +119,146 @@ export abstract class LendProvider<
...
+          marketId.address.toLowerCase() &&
+        allowedMarket.chainId === marketId.chainId,
+    )
+
+    if (!foundMarket) {
+      throw new Error(
+        `Market ${marketId.address} on chain ${marketId.chainId} is not in the market allowlist`,
+      )
+    }
+  }
```
tremarkley, line 195: hmm, why do we separate validating the network and the market allowlist? why not just have all validation be governed by the market allowlist. I think we could potentially simplify and do all validation through`validateConfigSupported` and delete `validateChainIdSupported` and `validateNetworkSupported`

packages/sdk/src/lend/provider.ts
```diff
@@ -50,63 +42,75 @@ export abstract class LendProvider<
...
    * Deposit assets to a market (alias for lend)
    * @param asset - Asset token address to deposit
    * @param amount - Amount to deposit (in wei)
+   * @param chainId - Chain ID for the transaction
    * @param marketId - Optional specific market ID
    * @param options - Optional deposit configuration
    * @returns Promise resolving to deposit transaction details
    */
-  abstract deposit(
+  async deposit(
```
tremarkley, line 82: we should delete this function soon, so we can avoid the new demo using it

packages/sdk/src/lend/provider.ts
```diff
@@ -115,19 +119,146 @@ export abstract class LendProvider<
...
+  protected abstract _getMarketBalance(
+    marketId: LendMarketId,
+    walletAddress: Address,
+  ): Promise<{
+    balance: bigint
+    balanceFormatted: string
+    shares: bigint
+    sharesFormatted: string
+    chainId: number
+  }>
```
tremarkley: nit: it would be good to use a shared type for this

packages/sdk/src/lend/providers/morpho/index.ts
```diff
@@ -19,27 +18,22 @@ import { LendProvider } from '../../provider.js'
...
-    name: 'Base Sepolia',
-    morphoAddress: '0xBBBBBbbBBb9cC5e90e3b3Af64bdAF62C37EEFFCb',
-  },
-} as const
+export const SUPPORTED_NETWORK_IDS = [
+  mainnet.id, // 1
+  unichain.id, // 130
+  base.id, // 8453
+  baseSepolia.id, // 84532
+] as const
```
tremarkley: this should be typed as `SUPPORTED_NETWORK_IDS`: `SupportedChainIds[]` that way it will not be possible to include a chain id that is not supported by the sdk

packages/sdk/src/lend/providers/morpho/index.ts
```diff
@@ -78,7 +72,7 @@ export class LendProviderMorpho extends LendProvider<MorphoLendConfig> {
       // 2. Get vault information for APY
       const vaultInfo = await this.getMarket({
         address: selectedVaultAddress,
-        chainId: chainId || (130 as SupportedChainId), // TODO: Get chain ID dynamically
+        chainId: chainId as SupportedChainId,
```
tremarkley: nit: it would be good to make it so that this cast isnt necessary so that we can ensure that the chainId is actually a `SupportedChainId`

packages/sdk/src/lend/providers/morpho/sdk.ts
```diff
@@ -141,18 +135,17 @@ export async function getVault(params: GetVaultParams): Promise<LendMarket> {
...
 
     return {
-      chainId: config.chainId,
+      chainId: marketId.chainId,
       address: marketId.address,
-      name: config.name,
-      asset: (config.asset.address[config.chainId] ||
-        Object.values(config.asset.address)[0]) as Address,
+      name: `Vault ${marketId.address}`, // Generic name since we don't have config metadata
+      asset: '0x' as Address, // TODO: Get asset address from vault data
```
tremarkley: would be good to double check that removing this doesnt break anything in the demo. I want to say that `wallet balance` or `wallet lend` may rely on this

packages/sdk/src/lend/namespaces/WalletLendNamespace.ts
its-everdred: it will, i keep nudging these others function along, but next week I'll be updating them all directly

packages/sdk/src/lend/provider.ts
its-everdred: yes! will do next week

packages/sdk/src/lend/provider.ts
its-everdred: let's consolidate to Chain

packages/sdk/src/lend/provider.ts
its-everdred: good catch, deleted

packages/sdk/src/lend/provider.ts
its-everdred: I deleted `validateNetworkSupported` and renamed`validateChainIdSupported` -> `validateProviderSupported`. 

the difference is:
`validateConfigSupported` == makes sure it's allowed by the dev's `VerbsConfig` (they don't have it blocked, or missing from an existing allow list)
`validateProviderSupported` == makes sure it's allowed by the provider (ex: ensure morpho has markets on that chain)

packages/sdk/src/lend/provider.ts
its-everdred: yup, will be removing soon

packages/sdk/src/lend/provider.ts
its-everdred: will do. ill probably make shared types for all the documented LendProvider function params in the next PRs

packages/sdk/src/lend/providers/morpho/index.ts
its-everdred: Morpho's sdk actually exports a `ChainId` enum with all of the networks they have markets on, so i just updated this to use their type.

packages/sdk/src/lend/provider.ts
tremarkley: yeah, I was just thinking this could be simplified. By doing the validation on the market allowlist + the lend provider when the lend provider is created and then just validate against the subset of the market allowlist for the given lend provider. 

The flow would be something like this:
1. When LendProvider is created create a LendProvider market allowlist which is a combo of the supported chains for the LendProvider plus the markets in the allowlist for the lend provider
2. Now whenever you need to validate that a given market/chain is valid you can just check that the Lend provider market allowlist contains that market

Your way works too though!

tremarkley (approved):

packages/sdk/src/lend/provider.ts
its-everdred: ohh i follow and like that. so essentially resolve and save the overlap list. that'll work for chainId, but probably not market Ids since we will soon pull markets from morpho's api as well. l'll confirm with Pierce about the intended specifics and include in a future PR

packages/sdk/src/lend/providers/morpho/index.ts
its-everdred: done

packages/sdk/src/lend/providers/morpho/sdk.ts
its-everdred: reverted this change, will refactor the way we handle this in a future PR

packages/sdk/src/lend/provider.ts
its-everdred: done

packages/sdk/src/lend/provider.ts
its-everdred: will do in this [issue](https://github.com/ethereum-optimism/verbs/issues/130)

PR #137, opened 2025-09-24
feat: add support for Turnkey server wallets

### LINKED ISSUE #139: Wallet adapter for Turnkey server wallets

tremarkley:
Closes https://github.com/ethereum-optimism/verbs/issues/139

## Description
Add `TurnKeyWallet` class for supporting Turnkey server wallets

### DIFF
--- a/packages/sdk/src/wallet/TurnkeyWallet.ts
+++ b/packages/sdk/src/wallet/TurnkeyWallet.ts
@@ -1,0 +1,94 @@
+import type { TurnkeySDKClientBase } from '@turnkey/core'
+import type { TurnkeyClient } from '@turnkey/http'
+import type { TurnkeyServerClient } from '@turnkey/sdk-server'
+import { createAccount } from '@turnkey/viem'
+import type { Address, LocalAccount, WalletClient } from 'viem'
+import { createWalletClient, fallback, http } from 'viem'
+
+import type { SupportedChainId } from '@/constants/supportedChains.js'
+import type { ChainManager } from '@/services/ChainManager.js'
+import { Wallet } from '@/wallet/base/Wallet.js'
+
+/**
+ * Turnkey wallet implementation
+ * @description Wallet implementation using Turnkey service
+ */
+export class TurnkeyWallet extends Wallet {
+  public address!: Address
+  public signer!: LocalAccount
+  /**
+   * Turnkey client instance (HTTP, server, or core SDK base)
+   */
+  private readonly client:
+    | TurnkeyClient
+    | TurnkeyServerClient
+    | TurnkeySDKClientBase
+  /**
+   * Turnkey organization ID that owns the signing key
+   */
+  private readonly organizationId: string
+  /**
+   * This can be a wallet account address, private key address, or private key ID.
+   */
+  private readonly signWith: string
+  /**
+   * Ethereum address to use for this account, in the case that a private key ID is used to sign.
+   * If left undefined, `createAccount` will fetch it from the Turnkey API.
+   * We recommend setting this if you're using a passkey client, so that your users are not prompted for a passkey signature just to fetch their address.
+   * You may leave this undefined if using an API key client.
+   */
+  private readonly ethereumAddress?: string
+
+  private constructor(params: {
+    chainManager: ChainManager
+    client: TurnkeyClient | TurnkeyServerClient | TurnkeySDKClientBase
+    organizationId: string
+    signWith: string
+    ethereumAddress?: string
+  }) {
+    const { chainManager, client, organizationId, signWith, ethereumAddress } =
+      params
+    super(chainManager)
+    this.client = client
+    this.organizationId = organizationId
+    this.signWith = signWith
+    this.ethereumAddress = ethereumAddress
+  }
+
+  static async create(params: {
+    chainManager: ChainManager
+    client: TurnkeyClient | TurnkeyServerClient | TurnkeySDKClientBase
+    organizationId: string
+    signWith: string
+    ethereumAddress?: string
+  }): Promise<TurnkeyWallet> {
+    const wallet = new TurnkeyWallet(params)
+    await wallet.initialize()
+    return wallet
+  }
+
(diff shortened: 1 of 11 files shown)


tremarkley:
* **#138** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/138?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#137** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/137?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/137?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pnpm-lock.yaml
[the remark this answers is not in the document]
tremarkley: #wiz_ignore this vulnerability is related to a peer dependency which is only installed in the dev build of this library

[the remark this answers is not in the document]
tremarkley: #wiz_ignore this vulnerability would cause the app server to potentially crash if exploited. the app server is only being used for a demo application and a crash would not impact any production services.

packages/sdk/src/wallet/providers/TurnkeyHostedWalletProvider.ts
```diff
@@ -0,0 +1,56 @@
...
+   * Convert a Turnkey hosted wallet context into a Verbs wallet
+   * @description Creates a `TurnkeyWallet` configured with the provider's Turnkey
+   * client and organization.
+   * @param params - Options for creating the Verbs wallet from Turnkey context
+   * @param params.signWith - Wallet account address, private key address, or private key ID
+   * @param params.ethereumAddress - Ethereum address to use for this account, in the case that a private key ID is used to sign.
+   * @returns Promise resolving to a Verbs-compatible wallet instance
+   */
+  async toVerbsWallet(
+    params: TurnkeyHostedWalletToVerbsWalletOptions,
```
its-everdred, line 46: I've been avoiding using this `params` attr, but I think it actually makes sense. I'll probably switch LendProvider functions to use it.

packages/sdk/src/wallet/TurnkeyWallet.ts
```diff
@@ -0,0 +1,94 @@
...
+   * Turnkey client instance (HTTP, server, or core SDK base)
+   */
+  private readonly client:
+    | TurnkeyClient
+    | TurnkeyServerClient
+    | TurnkeySDKClientBase
+  /**
+   * Turnkey organization ID that owns the signing key
+   */
+  private readonly organizationId: string
```
its-everdred, line 29: ❓ Will this ever change from wallet to wallet? If not, should it live at an abstraction above the wallet instance, like the `VerbsConfig`?

packages/sdk/src/wallet/TurnkeyWallet.ts
```diff
@@ -0,0 +1,94 @@
...
+  private constructor(params: {
+    chainManager: ChainManager
+    client: TurnkeyClient | TurnkeyServerClient | TurnkeySDKClientBase
+    organizationId: string
+    signWith: string
+    ethereumAddress?: string
+  }) {
+    const { chainManager, client, organizationId, signWith, ethereumAddress } =
+      params
+    super(chainManager)
```
its-everdred, line 51: 🤔 I know this isn't new to this class, but I've wondered if the ChainManager needs to be passed through every instance of everything, or if it can live at an abstraction above? Thinking out loud, I'm not sure what that would look like, but like supported assets and supported networks, it's not something that will change from wallet to wallet.

its-everdred (approved):
some ❓s but looks good

packages/sdk/src/wallet/providers/TurnkeyHostedWalletProvider.ts
tremarkley: yeah, I'm in general a fan of using param types, I find it more readable and ergonomic

packages/sdk/src/wallet/TurnkeyWallet.ts
tremarkley: yeah, so looking at the turnkey docs, it is possible that all wallets could be under the same `organizationId`, but they also have support for each wallet being in its own sub-organization, which would mean a different `organizationId` per wallet. Here is more info: https://docs.turnkey.com/embedded-wallets/sub-organizations-as-wallets.

Given this I think it makes sense to abstract this at the wallet-level for now and we can always pull it into a higher level later if we find this is a big point of friction, but at least this design supports both use-cases (ie wallets as sub-organizations or all wallets under same organization)

packages/sdk/src/wallet/TurnkeyWallet.ts
tremarkley: even though `ChainManager` doesnt change from wallet to wallet, it does change for each `Verbs` instance and the wallet needs a way to access it. Right now we've opted for accessing it through dependency injection, which has it's tradeoffs, but the only other way we could potentially access it is through a global variable, which might make things more difficult to test and harder to trace. Let me know if there is another way you were thinking though or if you think a global variable would be preferred here

its-everdred: Having thought through it a bit more, I think this is totally fine. Global state gets dangerous and passing around a single dependency is very manageable. 

I’m imaging a future where additional context providers are added to verbs that allow the dev to overwrite functionality like:
- `ChainExplorerProvider`:  overwrite all receipt urls with blockscout or etherscan
- `TransactionSimulationProvider`:  overwrite all wallet.action.quote functions with tenderly or alchemy simulations

If we get to that world, we might pull these dependencies into a shared context provider registry and pass that single object around, but that's well into the future.

tremarkley: makes sense! if we get to a point where providers are coupled, I could see combining them making sense

PR #153, opened 2025-09-30
LendProvider: Implement closePosition()

its-everdred:
## Problem

A user cannot close a lend position.

## Solution

Implement morpho `closePosition()`
close-position


packages/sdk/src/lend/providers/morpho/index.test.ts
```diff
@@ -70,20 +71,107 @@ describe('LendProviderMorpho', () => {
...
+                supplyShares: BigInt(1000000e6),
+                supplyAssets: BigInt(1000000e6),
+                market: {
+                  supplyApy: BigInt(3e16),
+                },
+              },
+            },
+          ],
+        ]),
+      }
```
its-everdred, line 95: going to export this from a `__mocks__` folder in the next PR

packages/demo/frontend/src/api/verbsApi.ts
```diff
@@ -186,16 +186,17 @@ class VerbsApiClient {
     })
   }
 
-  async getMarketBalance(
+  async getPosition(
```
its-everdred, line 189: forgot to rename this in the last PR

packages/sdk/src/lend/providers/morpho/index.ts
```diff
@@ -130,23 +119,51 @@ export class LendProviderMorpho extends LendProvider<MorphoLendConfig> {
...
+        timestamp: currentTimestamp,
+        transactionData: {
+          closePosition: {
+            to: params.marketId.address,
+            data: withdrawCallData,
+            value: 0n,
+          },
+        },
+        slippage: params.options?.slippage || this._config.defaultSlippage,
+      }
```
its-everdred, line 161: improving this return type in the next PR

packages/demo/backend/src/controllers/lend.ts
```diff
@@ -9,16 +9,34 @@ import { validateRequest } from '../helpers/validation.js'
 import * as lendService from '../services/lend.js'
 
 const OpenPositionRequestSchema = z.object({
+  params: z.object({
+    marketId: z
+      .string()
+      .regex(/^0x[a-fA-F0-9]{40}$/, 'Invalid vault address format'),
+    chainId: z.string().min(1, 'chainId is required'),
+  }),
```
tremarkley, line 17: nit: what is the reasoning behind having some of the request object in params and some inside the body?

packages/demo/backend/src/controllers/lend.ts
```diff
@@ -9,16 +9,34 @@ import { validateRequest } from '../helpers/validation.js'
...
+})
+
+const ClosePositionRequestSchema = z.object({
+  params: z.object({
+    marketId: z
       .string()
       .regex(/^0x[a-fA-F0-9]{40}$/, 'Invalid vault address format'),
+    chainId: z.string().min(1, 'chainId is required'),
+  }),
```
tremarkley, line 33: same question here re: splitting between `params` and `body`

packages/demo/backend/src/controllers/lend.ts
```diff
@@ -9,16 +9,34 @@ import { validateRequest } from '../helpers/validation.js'
 import * as lendService from '../services/lend.js'
 
 const OpenPositionRequestSchema = z.object({
+  params: z.object({
+    marketId: z
```
tremarkley, line 13: nit: should this be `marketAddress`? I think in some place we use `marketId` as a combination of the market address and the chain id, so this was a little confusing to me. You could also move this into the `body` and type it as `marketId: z.object({ address: z.string(), chainId: z.string() })`

packages/demo/backend/src/controllers/lend.ts
```diff
@@ -9,16 +9,34 @@ import { validateRequest } from '../helpers/validation.js'
...
       .string()
       .regex(/^0x[a-fA-F0-9]{40}$/, 'Invalid token address format'),
-    chainId: z.number().min(1, 'chainId is required'),
-    vaultAddress: z
+  }),
+})
+
+const ClosePositionRequestSchema = z.object({
+  params: z.object({
+    marketId: z
```
tremarkley, line 29: nit: same comment here re: `marketId` type

packages/demo/backend/src/services/lend.ts
```diff
@@ -112,63 +112,73 @@ export async function formatMarketBalanceResponse(
...
-}: OpenPositionParams): Promise<Hash> {
-  // Get wallet based on identifier type
-  const wallet = await getWallet(userId, isUserWallet)
+interface PositionResponse {
+  hash: string
+  blockExplorerUrl: string
+  amount: number
+  tokenAddress: Address
+  chainId: number
+  vaultAddress: Address
```
tremarkley, line 131: I thought we went with calling them markets and not vaults? or are there some scenarios we still want to refer to them as vaults?

packages/sdk/src/lend/namespaces/WalletLendNamespace.spec.ts
```diff
@@ -170,7 +169,7 @@ describe('WalletLendNamespace', () => {
         apy: 0.05,
         timestamp: Date.now(),
         transactionData: {
-          deposit: {
+          closePosition: {
             to: closeParams.marketId.address,
             value: 0n,
             data: '0x' as const,
```
tremarkley, line 189: nit: might be good to have a non-zero value here so the test verifies that the calldata is passed to `lendProvider.closePosition`

packages/sdk/src/lend/namespaces/WalletLendNamespace.ts
```diff
@@ -43,40 +43,35 @@ export class WalletLendNamespace<
    * @description Signs and sends a lend transaction from the wallet for the given amount and asset
    */
   async openPosition(params: LendOpenPositionParams): Promise<Hash> {
-    const lendOptions = {
-      ...params.options,
-      receiver: this.wallet.address,
-    }
-
     const lendTransaction = await this.provider.openPosition({
```
tremarkley, line 47: nit: I think we should rename this function. `openPosition` to me sounds like it is going to open the position but really what this is doing is creating the `calls` that then need to be executed in order to open the position.

packages/sdk/src/lend/namespaces/WalletLendNamespace.ts
```diff
@@ -43,40 +43,35 @@ export class WalletLendNamespace<
...
-    if (transactionData.approval) {
+    if (transactionData.approval && transactionData.openPosition) {
       return await this.wallet.sendBatch(
-        [transactionData.approval, transactionData.deposit],
+        [transactionData.approval, transactionData.openPosition],
         params.marketId.chainId,
       )
     }
 
+    if (!transactionData.openPosition) {
```
tremarkley, line 70: why is it even possible for `this.provider.openPosition` to return an undefined `transactionData.openPosition`? should we just have it so it's always defined and if for some reason it's not defined, then `this.provider.openPosition` throws an error? wdyt?

packages/sdk/src/lend/namespaces/WalletLendNamespace.ts
```diff
@@ -102,16 +97,9 @@ export class WalletLendNamespace<
    * @returns Promise resolving to transaction hash
    */
   async closePosition(params: ClosePositionParams): Promise<Hash> {
-    const closeOptions = {
-      ...params.options,
-      receiver: this.wallet.address,
-    }
-
     const closeTransaction = await this.provider.closePosition({
```
tremarkley, line 107: same comment here re: naming.

packages/sdk/src/lend/namespaces/WalletLendNamespace.ts
```diff
@@ -127,15 +115,12 @@ export class WalletLendNamespace<
       )
     }
 
-    if (transactionData.approval) {
-      return await this.wallet.sendBatch(
-        [transactionData.approval, transactionData.deposit],
-        params.marketId.chainId,
-      )
+    if (!transactionData.closePosition) {
```
tremarkley, line 125: same comment here around making sure `transactionData.closePosition)` is always defined.

packages/sdk/src/lend/provider.ts
```diff
@@ -66,6 +66,10 @@ export abstract class LendProvider<
    * @returns Promise resolving to lending transaction details
    */
   async openPosition(params: LendOpenPositionParams): Promise<LendTransaction> {
+    if (!params.walletAddress) {
```
tremarkley, line 69: should we just make `walletAddress` a required param in `LendOpenPositionParams` or is there a case where it should be undefined?

packages/sdk/src/lend/provider.ts
```diff
@@ -155,34 +160,43 @@ export abstract class LendProvider<
    * @param amount - Amount to withdraw (human-readable number)
    * @param asset - Asset to withdraw (optional, validated against marketId)
    * @param marketId - Market identifier containing address and chainId
+   * @param walletAddress - Wallet address for receiving assets and as owner
    * @param options - Optional withdrawal configuration
    * @returns Promise resolving to withdrawal transaction details
    */
   async closePosition(params: ClosePositionParams): Promise<LendTransaction> {
+    if (!params.walletAddress) {
```
tremarkley, line 168: same comment here around making `walletAddress` a required param

packages/sdk/src/lend/provider.ts
```diff
@@ -155,34 +160,43 @@ export abstract class LendProvider<
...
       chainId: params.marketId.chainId,
     })
 
-    // If asset is provided, validate it matches the market's asset
     if (params.asset) {
       validateMarketAsset(market, params.asset)
     }
 
-    const assetAddress = market.asset as Address
+    const assetMetadata = params.asset?.metadata
```
tremarkley, line 184: similar commenet here, why not just have this be a required param?

packages/sdk/src/lend/providers/morpho/index.ts
```diff
@@ -75,26 +75,17 @@ export class LendProviderMorpho extends LendProvider<MorphoLendConfig> {
...
-      if (!receiver) {
-        throw new Error(
-          'Receiver address is required for Morpho deposit operation',
-        )
-      }
-      const depositCallData = MetaMorphoAction.deposit(
-        params.amountWei,
-        receiver,
-      )
+      // MetaMorphoAction.deposit(assets, receiver)
```
tremarkley: nit: remove this

packages/sdk/src/lend/providers/morpho/index.ts
```diff
@@ -130,23 +119,54 @@ export class LendProviderMorpho extends LendProvider<MorphoLendConfig> {
...
+      const assets = params.amount
+      const receiver = params.walletAddress
+      const owner = params.walletAddress
+      const withdrawCallData = MetaMorphoAction.withdraw(
+        assets,
+        receiver,
+        owner,
+      )
+
+      const currentTimestamp = Math.floor(Date.now() / 1000)
```
tremarkley, line 145: why do we need to return the timestamp?

tremarkley (approved):

packages/demo/backend/src/controllers/lend.ts
its-everdred: in this request, market address and chain ID could also be totally fine in the body. since the backend has specific configured markets and chains that it supports, it felt like the market position is sort of like the RESTful resource of the request, with a create (openPosition) and an update/delete (closePosition). but that's probably a stretch.

packages/demo/backend/src/controllers/lend.ts
its-everdred: good catch, I want to stick with `marketId: { address, chainId }` . I'll update

packages/demo/backend/src/controllers/lend.ts
its-everdred: for simplicity, im going to move everything into the body

its-everdred: Ill move the marketId info to the body for the POST requests for simplicity

packages/sdk/src/lend/providers/morpho/index.ts
its-everdred: this is left over and can be removed.

its-everdred (commented):
the next branch has a lot of lend refactors, going to implement all requested changes in the next PR

PR #158, opened 2025-10-01
feat: add `createSigner` to wallet namespace

### LINKED ISSUE #159: add createSigner to wallet namespace

tremarkley:
Closes https://github.com/ethereum-optimism/verbs/issues/159

Exposes a `createSigner` method under `HostedWalletProvider` and `WalletNamespace`. This simplifies the flow for when a signer is needed for a `SmartWallet` instance

### DIFF
--- a/packages/sdk/src/wallet/core/namespace/__tests__/WalletNamespace.spec.ts
+++ b/packages/sdk/src/wallet/core/namespace/__tests__/WalletNamespace.spec.ts
@@ -1,4 +1,5 @@
 import type { PrivyClient } from '@privy-io/server-auth'
+import { getAddress } from 'viem'
 import { unichain } from 'viem/chains'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
@@ -9,7 +10,7 @@ import { createMockPrivyClient } from '@/test/MockPrivyClient.js'
 import { getRandomAddress } from '@/test/utils.js'
 import { WalletNamespace } from '@/wallet/core/namespace/WalletNamespace.js'
 import { DefaultSmartWalletProvider } from '@/wallet/core/providers/smart/default/DefaultSmartWalletProvider.js'
-import { WalletProvider } from '@/wallet/core/providers/wallet/WalletProvider.js'
+import { WalletProvider } from '@/wallet/core/providers/WalletProvider.js'
 import { Wallet } from '@/wallet/core/wallets/abstract/Wallet.js'
 import { DefaultSmartWallet } from '@/wallet/core/wallets/smart/default/DefaultSmartWallet.js'
 import { PrivyHostedWalletProvider } from '@/wallet/node/providers/hosted/privy/PrivyHostedWalletProvider.js'
@@ -97,7 +98,7 @@ describe('WalletNamespace', () => {
       const hostedWallet =
         await walletProvider.hostedWalletProvider.toVerbsWallet({
           walletId: privyWallet.id,
-          address: privyWallet.address,
+          address: getAddress(privyWallet.address),
         })
       const owners = [getRandomAddress(), hostedWallet.address]
       const nonce = BigInt(123)
@@ -145,7 +146,7 @@ describe('WalletNamespace', () => {
       const hostedWallet =
         await walletProvider.hostedWalletProvider.toVerbsWallet({
           walletId: privyWallet.id,
-          address: privyWallet.address,
+          address: getAddress(privyWallet.address),
         })
       const deploymentOwners = [hostedWallet.address, getRandomAddress()]
       const signerOwnerIndex = 0
@@ -188,7 +189,7 @@ describe('WalletNamespace', () => {
       const hostedWallet =
         await walletProvider.hostedWalletProvider.toVerbsWallet({
           walletId: privyWallet.id,
-          address: privyWallet.address,
+          address: getAddress(privyWallet.address),
         })
 
       await expect(
@@ -224,7 +225,7 @@ describe('WalletNamespace', () => {
       const hostedWallet =
         await walletProvider.hostedWalletProvider.toVerbsWallet({
           walletId: privyWallet.id,
-          address: privyWallet.address,
+          address: getAddress(privyWallet.address),
         })
       const toVerbsWalletSpy = vi.spyOn(
         walletProvider.hostedWalletProvider,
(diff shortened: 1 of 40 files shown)


tremarkley:
* **#158** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/158?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/sdk/src/wallet/node/providers/hosted/turnkey/TurnkeyHostedWalletProvider.ts
```diff
@@ -22,15 +24,13 @@ export class TurnkeyHostedWalletProvider extends HostedWalletProvider<
...
    * @param client - Turnkey client instance (HTTP, server, or core SDK base)
-   * @param organizationId - Turnkey organization ID that owns the signing key
    * @param chainManager - Chain manager used to resolve chains and RPC transports
    */
   constructor(
     private readonly client:
       | TurnkeyHttpClient
       | TurnkeyServerClient
       | TurnkeySDKClientBase,
-    private readonly organizationId: string,
```
tremarkley, line 33: I realized that `organizationId` should actually be provided at wallet creation and not at verbs creation so removed it from the `WalletProvider` constructor and into the `toVerbsWallet` arguments

packages/sdk/src/wallet/core/namespace/__tests__/WalletNamespace.spec.ts
```diff
@@ -97,7 +98,7 @@ describe('WalletNamespace', () => {
       const hostedWallet =
         await walletProvider.hostedWalletProvider.toVerbsWallet({
           walletId: privyWallet.id,
-          address: privyWallet.address,
+          address: privyWallet.address as Address,
```
its-everdred: ❓ Can we remove the need for this? If it's a hassle, we should take a full repo pass removing type casting later.

tremarkley: type casts in tests arent the worst thing but we should definitely try to remove as many of them in non test files as possible. I updated this file to use `getAddress` instead of `as Address`

packages/sdk/src/wallet/node/providers/hosted/turnkey/TurnkeyHostedWalletProvider.ts
its-everdred: why is that? It feels weird that the dev would need to keep passing a static ID for every wallet creation

tremarkley: it's not a static id. depending on how the user has their Turnkey organization set up there might be a different organization id per wallet. Turnkey has a hierarchical organization structure.  More info [here](https://docs.turnkey.com/embedded-wallets/sub-organizations-as-wallets#sub-organizations-as-end-user-controlled-wallets)

packages/sdk/src/wallet/node/wallets/hosted/turnkey/TurnkeyWallet.ts
```diff
@@ -84,7 +84,7 @@ export class TurnkeyWallet extends Wallet {
   }
 
   private async createAccount(): Promise<LocalAccount> {
```
its-everdred: Is there a reason to keep this differentiation?

tremarkley: oh yeah, it probably makes sense to rename this to `createSigner` as well

packages/sdk/src/wallet/react/wallets/hosted/dynamic/DynamicWallet.ts
```diff
@@ -83,22 +81,8 @@ export class DynamicWallet extends Wallet {
    * @throws Error if wallet retrieval fails or signing operations are not supported
    */
   private async createAccount(): Promise<LocalAccount> {
```
its-everdred: Same question about differentiating `createAccount` vs `createSigner`

tremarkley: good catch, just pushed up a commit, renaming all of these from `createAccount` -> `createSigner`

packages/sdk/src/wallet/node/providers/hosted/turnkey/TurnkeyHostedWalletProvider.ts
its-everdred: 👍 ah interesting, good call making it Turnkey specific then

packages/sdk/src/wallet/react/wallets/hosted/privy/PrivyWallet.ts
```diff
@@ -72,16 +71,8 @@ export class PrivyWallet extends Wallet {
    * @throws Error if wallet retrieval fails or signing operations are not supported
    */
   private async createAccount(): Promise<LocalAccount> {
```
its-everdred, line 73: `createAccount()` -> `createSigner()`

packages/sdk/src/wallet/node/providers/hosted/turnkey/TurnkeyHostedWalletProvider.ts
its-everdred: would it be a hassle to make it optional on the `createSigner()` and default to a value in the `VerbsConfig` if not provided? Considering the other providers don't even have sub-organizations, I imagine _most_ simple use cases will have a single org. It's weird to have the required extra param for the edge-case complex org.

tremarkley: hmm, I actually dont think it's an edge case. By default on the frontend each wallet thats created automatically gets put in a different organization. And turnkey even limits you to only creating 100 wallets per organization and encourages you to use sub-organizations:

> custodial wallets should still create segregated sub-organizations for their end-users to avoid limits (we currently have a maximum of 100 users per organization, whereas an organization can have unlimited sub-organizations).

if anything the edge case would be the same org id for every wallet. By passing in organization id, this at least supports both use cases and I dont think it's a massive hassle to have to pass in a static value on each call. I would prefer holding off on adding the extra code to support an optional organization id only if we find this is a common use case, otherwise I think it just makes this code more confusing.

its-everdred: oh wow okay lets keep this. thanks for explaining!

its-everdred (approved):
👍 Looks good. I was trying to think of a way reduce the boilerplate needed to add new functions, but there's not a lot to remove. Most of this PR is mapping a provider specific `createSigner()` function to a provider specific util `createSigner()`, so you could remove some by moving the util function logic into the provider classes, but I don't feel strongly about that, especially if we plan to start using these util files for more.

PR #161, opened 2025-10-02
feat: add support for non-lazy deployments of `DefaultSmartWallet`

tremarkley:
Part of https://github.com/ethereum-optimism/verbs/issues/98

### Motivation

Previously, smart wallets only supported lazy deployment via ERC-4337 initCode, which meant deployments happened implicitly during the first transaction. This worked for simple use cases, but developers need more control over when and where wallets are deployed, especially for multi-chain scenarios.

### Changes
- Added `deploy()` method to `DefaultSmartWallet` for explicit wallet deployment
- ` createSmartWallet()` now calls `deploy()` and returns `SmartWalletDeploymentResult` containing:
  - `wallet`: The created `SmartWallet` instance
  - `deploymentSuccesses`: Array of successful deployments per chain
  - `deploymentFailures`: Array of failed deployments with error details
- Added optional `deploymentChainIds` parameter to control which chains to deploy to (pass `[]` to have lazy deployment)

### DIFF
--- a/packages/demo/backend/src/services/wallet.spec.ts
+++ b/packages/demo/backend/src/services/wallet.spec.ts
@@ -54,12 +54,15 @@ describe('Wallet Service', () => {
       mockPrivyClient.walletApi.createWallet.mockResolvedValue(mockPrivyWallet)
 
       const mockWallet = {
-        id: 'wallet-123',
-        address: '0x1234567890123456789012345678901234567890',
-        signer: {
+        wallet: {
+          id: 'wallet-123',
           address: '0x1234567890123456789012345678901234567890',
+          signer: {
+            address: '0x1234567890123456789012345678901234567890',
+          },
+          lend: {},
         },
-        lend: {},
+        deployments: [{ chainId: 1, receipt: undefined, success: true }],
       }
 
       mockVerbs.wallet.createSmartWallet.mockResolvedValue(mockWallet)
@@ -113,6 +116,7 @@ describe('Wallet Service', () => {
           type: 'local',
           address: '0x1234567890123456789012345678901234567890',
         },
+        owners: ['0x1234567890123456789012345678901234567890'],
         deploymentOwners: ['0x1234567890123456789012345678901234567890'],
       })
       expect(result).toEqual(mockWallet)
@@ -271,6 +275,7 @@ describe('Wallet Service', () => {
           address: '0x1234567890123456789012345678901234567890',
           type: 'local',
         },
+        owners: ['0x1234567890123456789012345678901234567890'],
         deploymentOwners: ['0x1234567890123456789012345678901234567890'],
       })
       expect(mockWallet.getBalance).toHaveBeenCalled()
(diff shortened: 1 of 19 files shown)


tremarkley:
* **#161** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/verbs/161?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/demo/backend/src/services/wallet.spec.ts
```diff
@@ -50,12 +50,15 @@ describe('Wallet Service', () => {
...
+        wallet: {
+          id: 'wallet-123',
           address: '0x1234567890123456789012345678901234567890',
+          signer: {
+            address: '0x1234567890123456789012345678901234567890',
+          },
+          lend: {},
         },
-        lend: {},
+        deploymentSuccesses: [{ chainId: 1, receipt: undefined }],
```
its-everdred: I'm curious why `deploymentSuccesses` + `deploymentFailures` vs `deployments: [ { chainId, status, receipt?, error? } ]`

packages/sdk/src/wallet/core/providers/smart/default/DefaultSmartWalletProvider.ts
```diff
@@ -50,27 +53,67 @@ export class DefaultSmartWalletProvider extends SmartWalletProvider {
...
       owners,
       signer,
       chainManager: this.chainManager,
       lendProvider: this.lendProvider,
       nonce,
       attributionSuffix: this.attributionSuffix,
     })
+
+    const deploymentResults = await Promise.allSettled(
+      (deploymentChainIds ?? this.chainManager.getSupportedChains()).map(
```
its-everdred, line 86: ❓ We don't need to figure it out in this PR, but how do we want to handle (on encourage devs to handle) this scenario:
1. VerbsConfig.chains == [1,2,3]
2. deploy some wallets
3. dev updates VerbsConfig.chains = [2,3,4] // removed + adds chain support. new wallet deployments needed

Comes with the challenge of building a stateless library that deals with a lot of state I guess.

its-everdred (approved):

packages/demo/backend/src/services/wallet.spec.ts
tremarkley: yeah, good call no need to split them out and complicate the schema. I updated this to just be `deployments: [ { chainId, status, receipt?, error? } ]`

packages/sdk/src/wallet/core/providers/smart/default/DefaultSmartWalletProvider.ts
tremarkley: yeah, doing this in a stateless library is tricky. One thing that will somewhat help with this is that in the `getSmartWallet` function (which should be used by devs when they expect the wallet to already be deployed), the developer is expected pass in the `deploymentAddress` of the wallet.  If the  `deploymentAddress` they pass in does not correspond to a deployed smart wallet (or the address that the wallet would be lazily deployed at), then their txs will fail. This is not a great experience though, because the error isnt thrown when fetching the wallet, it would be thrown when they attempt to make a transaction with said wallet.

tremarkley: will think through this edge case a bit more

PR #164, opened 2025-10-03
Add Netlify toml

its-everdred:
## Problem

The `_redirects` file is currently only redirecting the sub-pages of `verbs.money` -> `actions.money`, but not the root domain itself. It simply renders the site with no redirect.

## Solution

We can solve this by adding a `!` to the end of the existing line in the `_redirects` file, but netlify also supports a more explicit syntax via a `netlify.toml`.


tremarkley (approved):

its-everdred:
no longer needed

PR #169, opened 2025-10-03
demo: create layout for new demo page

tarunkhasnavis:
Summary

  - Added new demo page layout with Earn, Info, and ActivityLog components
  - Updated frontend environment configuration with additional backend variables
  - Added Optimism and Privy branding assets


latest-faucet-deployment 2.json
tremarkley, line 1: this should be deleted. it's only meant for local development. it's supposed to be gitignored, but it looks like a `2` was added to the end of the name so it didnt get ignored

packages/demo/frontend/.env.example
tremarkley, line 1: we still want this file in here, but just with empty/dummy values. it's meant to be an example file that you can copy and then fill in the actual values

packages/demo/frontend/src/components/Info.tsx
```diff
@@ -0,0 +1,113 @@
...
+          className="text-2xl font-semibold mb-6"
+          style={{ color: '#1a1b1e' }}
+        >
+          What's Happening Under the Hood
+        </h2>
+
+        <ul className="space-y-3 mb-8" style={{ color: '#000000', fontSize: '14px' }}>
+          <li className="flex items-start">
+            <span className="mr-2">•</span>
+            <span>Verbs creates a smart wallet for you.</span>
```
tremarkley: nit: can you update anywhere that the string `Verbs` is being used to `Actions`?

packages/demo/frontend/src/components/Info.tsx
```diff
@@ -0,0 +1,113 @@
...
+          </li>
+          <li className="flex items-start">
+            <span className="mr-2">•</span>
+            <span>When you click Lend USDC, Verbs: gets a quote and opens a lending position on Base Sepolia.</span>
+          </li>
+        </ul>
+
+        <div className="space-y-3">
+          <a
+            href="https://github.com/ethereum-optimism/verbs"
```
tremarkley:
```suggestion
            href="https://github.com/ethereum-optimism/actions"
```

tremarkley (approved):

latest-faucet-deployment 2.json
tarunkhasnavis: fixed thanks

PR #170, opened 2025-10-03
demo: Add login flow and some more scaffolding for activity log

tarunkhasnavis:
- Added the privy wallet login flow 
- Enhanced activity logging with detailed item components


packages/demo/frontend/src/components/Earn.tsx
```diff
@@ -19,6 +34,109 @@ function Earn() {
...
+          // Use user ID as the wallet identifier
+          const userId = user.id
+
+          // Try to fetch balance first (wallet might already exist)
+          await fetchBalance(userId)
+          setWalletCreated(true)
+        } catch (error) {
+          console.error('Error initializing wallet:', error)
+        }
+      }
```
tremarkley, line 86: where is this creating the wallet?

packages/demo/frontend/src/components/Revolut.tsx
```diff
@@ -0,0 +1,987 @@
...
+      console.error('Failed to initialize real wallet:', error)
+      // Fallback to mock wallet only if API completely fails
+      const mockWallet: WalletData = {
+        id: 'revolut-demo-user-fallback',
+        address: '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266' as `0x${string}`
+      }
+      console.log('Using fallback mock wallet:', mockWallet)
+      setSelectedWallet(mockWallet)
+    }
+  }
```
tremarkley: instead of creating the wallet manually, we should just use the wallet that is associated with the user's privy account. Here is an example: https://github.com/ethereum-optimism/actions/blob/72f65200dd06816cb22ebece5a16af968cbbec8f/packages/demo/frontend/src/components/Terminal.tsx#L169

packages/demo/frontend/src/components/Revolut.tsx
```diff
@@ -0,0 +1,987 @@
...
+        }
+      }
+      
+      if (ethBalance >= 0.1 && usdcBalance >= 100) {
+        console.log('Wallet has sufficient balance')
+      }
+    } catch (error) {
+      console.error('Failed to check/fund wallet:', error)
+    }
+  }
```
tremarkley: the demo will be using a paymaster so we shouldnt need to fund the wallet with eth.

packages/demo/frontend/src/components/Revolut.tsx
```diff
@@ -0,0 +1,987 @@
...
+        // Convert CreateWalletResponse to WalletData format
+        const newWallet: WalletData = {
+          id: createResponse.userId,
+          address: createResponse.address
+        }
+        console.log('New wallet created:', newWallet)
+        setSelectedWallet(newWallet)
+        
+        // Fund the new wallet
+        await fundWalletIfNeeded(newWallet.id)
```
tremarkley: isnt the funding only supposed to happen when they click the Get USDC button?
Screenshot 2025-10-06 at 11 41 50 AM

packages/demo/frontend/src/components/Revolut.tsx
tremarkley, line 1: nit: why is this component named Revolut? we should keep it generic and not vendor specific

packages/demo/frontend/src/components/Revolut.tsx
```diff
@@ -0,0 +1,987 @@
...
+                    </div>
+                  </div>
+                  <div className="text-green-400 font-medium">
+                    +${transaction.amount.toFixed(2)}
+                  </div>
+                </div>
+              ))}
+            </div>
+          </div>
+        )}
```
tremarkley: this is a pretty big component. I know that the original demo app was pretty much built inside one massive component, but that was because it was meant to just be a quick poc that we threw away. if we want to maintain this, we should try and break up this file into multiple components

packages/demo/frontend/src/components/Venmo.tsx
```diff
@@ -0,0 +1,392 @@
...
+  }
+  totalAssets: string
+  totalShares: string
+  fee: number
+  owner: string
+  curator: string
+  lastUpdate: number
+}
+
+function Venmo() {
```
tremarkley: same thing here around naming. I think we should not have a vendor specific name here. Also, is part of the demo to use venmo? I dont think I saw that in the mocks

packages/demo/frontend/src/components/Venmo.tsx
```diff
@@ -0,0 +1,392 @@
...
+      // sum the total balance of all usdc tokens
+      const usdcBalance = usdcToken.reduce((acc, token) => acc + parseFloat(token.totalBalance) / 1e6, 0)
+
+      // Format balance to 2 decimal places
+      setWalletBalance(usdcBalance.toFixed(2))
+    } catch (error) {
+      console.error('Failed to fetch wallet balance:', error)
+      setWalletBalance('0.00')
+    }
+  }
```
tremarkley: this needs to be updated to read from the user's wallet and not the first wallet in the getAllWallets call

packages/demo/frontend/src/components/Venmo.tsx
```diff
@@ -0,0 +1,392 @@
...
+      const usdcBalance = usdcToken.reduce((acc, token) => acc + parseFloat(token.totalBalance) / 1e6, 0)
+
+      if (usdcBalance <= 0) {
+        // Fund wallet with USDC
+        await verbsApi.fundWallet(selectedWallet.id)
+        console.log('Wallet funded with USDC')
+      }
+    } catch (error) {
+      console.error('Failed to fund wallet:', error)
+    }
```
tremarkley: hmm, why is this in two places? do we fund from multiple places in the app? I left a similar comment but I believe the funding should happen when the get usdc button is clicked. If possible maybe we could turn that button into a component and then anywhere there needs to be a funding step, we can just use that component.

packages/demo/frontend/src/components/Venmo.tsx
```diff
@@ -0,0 +1,392 @@
...
+
+      if (ethBalance <= 0) {
+        // Fund wallet with ETH
+        await verbsApi.fundWallet(selectedWallet.id, 'ETH')
+        console.log('Wallet funded with ETH')
+      }
+    } catch (error) {
+      console.error('Failed to fund wallet:', error)
+    }
+  }
```
tremarkley: similar comment here as I left in Revolut.tsx. We actually shouldnt need an ETH funding step. The wallets in the demo will be hooked into a paymaster so the paymaster will pay the gas for any transactions done by the wallets.

packages/sdk/src/lend/providers/morpho/lend.supersim.test.ts
tremarkley, line 1: these tests should be deleted. We removed them, and I think you may have accidentally brought them back in

packages/sdk/src/lend/providers/morpho/vaults.ts
tremarkley, line 1: remove these changes

tremarkley (changes requested):

packages/demo/frontend/src/components/Revolut.tsx
tarunkhasnavis: this was created for our hackathon demos, removing them

packages/demo/frontend/src/components/Revolut.tsx
tarunkhasnavis: correct this was old code associated with the revolut demo, it is removed

packages/demo/frontend/src/components/Revolut.tsx
tarunkhasnavis: This revolut file is actually from the hackathon demo and was accidentally committed, I'm deleting it all together

packages/demo/frontend/src/components/Revolut.tsx
tarunkhasnavis: old revolut stuff, deleting

packages/demo/frontend/src/components/Venmo.tsx
tarunkhasnavis: Venmo demo - needs to be deleted

packages/sdk/src/lend/providers/morpho/lend.supersim.test.ts
tarunkhasnavis: removed

packages/sdk/src/lend/providers/morpho/vaults.ts
tarunkhasnavis: removed

packages/demo/frontend/src/components/Venmo.tsx
tarunkhasnavis: Sorry this is the Venmo app from the hackathon that accidentally got comitted, removing this

packages/demo/frontend/src/components/Earn.tsx
tarunkhasnavis: @tremarkley  its currently calling `actionsApi.createWallet()` but this is going to change with my next PR because I am creating an activity logger to wrap all actionsAPI calls

packages/demo/frontend/src/components/Earn.tsx
```diff
@@ -19,6 +34,117 @@ function Earn() {
...
+    }
+
+    initializeWallet()
+  }, [authenticated, user?.id, walletCreated, fetchBalance, getAuthHeaders])
```
tremarkley, line 90: you should probably just remove this, since the actual demo wont be using it

tremarkley (approved):

packages/demo/frontend/src/components/Earn.tsx
tarunkhasnavis: cool will do

PR #186, opened 2025-10-09
Demo improvements 2

its-everdred:
- [x] Remove unnecessary getMarket re-renders
- [x] Round down pennies to prevent an edge case causing tx failure
- [x] Add missing Activity logs
- [x] Add loading shimmer 
- [x] Update market styling
- [x] Add tx failure logging in backend


packages/demo/frontend/src/components/Action.tsx
```diff
@@ -114,12 +129,23 @@ function Action({ usdcBalance, isLoadingBalance, onMintUSDC, onTransactionSucces
       // Get the first transaction hash if available, or use userOpHash for account abstraction
       const txHash = result.transaction.transactionHashes?.[0] || result.transaction.userOpHash
       setTransactionHash(txHash)
+
+      // Use the block explorer URL from the backend (first one in the array)
+      const explorerUrl = result.transaction.blockExplorerUrls?.[0]
```
tremarkley, line 134: if there was an approval and lend that happened, the first transaction hash will be the approval. we should either show the last tx hash or render a list of the tx hash urls

tremarkley (approved):

packages/demo/frontend/src/components/Action.tsx
its-everdred: oh good catch, ill update this in the next pass!

PR #222, opened 2025-10-28
Add new home page summaries

its-everdred:
## Problem 

We use walls of code to explain the sdk without much in terms of visualization

## Solution
Desktop:
actions
Mobile:
mobile


packages/demo/frontend/src/pages/Docs.tsx
```diff
@@ -0,0 +1,36 @@
...
+          >
+            Docs
+          </h2>
+          <p
+            className="text-sm mb-12"
+            style={{ color: '#999', fontStyle: 'italic' }}
+          >
+            ⚠️ Our full docs are under construction. For now, these code
+            snippets offer a high level understanding of the sdk.
+          </p>
```
its-everdred, line 25: I added a temporary Docs page with our existing getting started content. Once pages in docs repo, we can delete this page and redirect

packages/demo/frontend/src/App.tsx
```diff
@@ -1,16 +1,26 @@
...
+import Docs from '@/pages/Docs'
+import { ROUTES } from '@/constants/routes'
 
 function App() {
+  // Prevent default scroll restoration to allow manual hash navigation
+  useEffect(() => {
+    if ('scrollRestoration' in window.history) {
+      window.history.scrollRestoration = 'manual'
+    }
+  }, [])
```
tremarkley: is this something we want to do application wide? will this potentially have unintended side effects in other parts of the demo app? I dont have a ton of context on what `window.history.scrollRestoration = 'manual'` does, so maybe it's not a big deal.

tremarkley (approved):

packages/demo/frontend/src/App.tsx
its-everdred: good catch, i'll make this home page specific

Issue #208, opened 2025-10-23
Home + Demo UX improvements


its-everdred:
## Home page 
- [ ] Consolidate pay vs send vs transfer wording. 
- [ ] Add section about frontend/backend with diagrams
- [ ] Split out docs into docs pages, create new simplified home page sections.
- [ ] Check “install” links and make sure front/back links to correct pages
- [ ] More detail explaining the smart wallet offering, consider diagram about this + signers, and where verbs sits in the stack.
- [ ] Once a user scrolls into the slide section, it appends an ID like #wallets to their url. when they leave the slide section, remove that ID from their url.
- [ ] Add a little more padding/margin between the content slide images.
- [ ] const embeddedWallet = primaryWallet is redundant in dynamic example
- [ ] Add semicolons to all code snippets
- [ ]  We should mention EVM at some point in the home page overview

## Demo

- [x] When hovering an activity log, it should highlight the data in the form that was fetched with it. Ex, Get Market should add some order of highlight box around the market data, get Position highlights the user's balance, get balance highlights user wallet balance, etc.
- [x] Add on-hover to red/orange/green status indicators to under the hood. "Transaction pending, succeeded, failed" or "Awaiting request, Request failed, Request succeeded" etc.
- [x] Change “now earning interest” on withdraw modal to something like "USDC received"
- [x] in Under the Hood, wallet.fund is not a real function of the sdk and should be changed to a description like "Receive demo funds". make it a wallet label
- [x] Add dynamic auth form X button to go back.
- [x] Update What’s Happening? Text to explain more clearly
- [x] Historic transactions should be saved specific to the wallet provider (currently seeing old TXs from another provider)
- [x] Fix duplicate Get Position activity logs
- [x] When vertically small browser, minimize What’s happening
- [ ] Create wallet doesn’t appear in activity log when the user first logs in. Figure out why and then save this event in local storage the way we save transaction events.
- [ ] Get wallet balance request shows red fail on first login.

Done:
- [x] Remove “hosted” from functions and comments on Home page code
- [x] Remove USDC from Gauntlet USDC in Market Config
- [x] Consider a tooltip for slippage
- [x] Reduce final line to remove the horizontal scroll in configure assets
- [x] Rename “MorphoMarket” -> “ExampleMarket”
- [x] After moving to docs, move npm install to the top.
- [x] Add ticket for send ENS support.
- [x] Bug where dynamic mint failed.

To sort:
[...]

PR #229, opened 2025-11-04
Demo UX improvements

its-everdred:
## Problem

User feedback found some [issues](https://github.com/ethereum-optimism/actions/issues/208) with the demo.

## Solution 

- [x] When hovering an activity log, it should highlight the data in the form that was fetched with it. Ex, Get Market should add some order of highlight box around the market data, get Position highlights the user's balance, get balance highlights user wallet balance, etc.
- [x] Add on-hover to red/orange/green status indicators to under the hood. "Transaction pending, succeeded, failed" or "Awaiting request, Request failed, Request succeeded" etc.
- [x] Change “now earning interest” on withdraw modal to something like "USDC received"
- [x] in Under the Hood, wallet.fund is not a real function of the sdk and should be changed to a description like "Receive demo funds". make it a wallet label
- [x] Add dynamic auth form X button to go back.
- [x] Update What’s Happening? Text to explain more clearly
- [x] Historic transactions should be saved specific to the wallet provider (currently seeing old TXs from another provider)
- [x] Fix duplicate Get Position activity logs
- [x] When vertically small browser, minimize What’s happening


packages/demo/frontend/.env.example
```diff
@@ -1,6 +1,6 @@
 # packages/demo/frontend/.env
 /** Required */
-VITE_VERBS_API_URL=
+VITE_ACTIONS_API_URL=http://localhost:3000
```
its-everdred, line 3: safe and accurate to include

tremarkley (approved):
nice!

PR #252, opened 2025-12-11
Add analytics

its-everdred:


packages/demo/frontend/src/utils/analytics.ts
```diff
@@ -0,0 +1,49 @@
+const MIXPANEL_TOKEN = '28b4162f2bd7a18d8006b0622ddd03e6'
```
tremarkley: nit: this is a config item, so we should make this an env var

packages/demo/frontend/src/utils/analytics.ts
```diff
@@ -0,0 +1,49 @@
...
+      persistence: 'localStorage',
+    })
+  }, 0)
+}
+
+export const trackEvent = (
+  eventName: string,
+  properties?: Record<string, unknown>,
+) => {
+  window.mixpanel?.track(eventName, properties)
```
tremarkley: once you do the above you can use the library instead of calling window https://github.com/ethereum-optimism/ecosystem-private/blob/main/apps/dev-console-interface/src/event-tracking/mixpanel.tsx

packages/demo/frontend/src/utils/analytics.ts
```diff
@@ -0,0 +1,49 @@
...
+  document.head.appendChild(script)
+
+  // Initialize after snippet loads
+  setTimeout(() => {
+    window.mixpanel?.init(MIXPANEL_TOKEN, {
+      track_pageview: true,
+      persistence: 'localStorage',
+    })
+  }, 0)
+}
```
tremarkley, line 14: I think it would be cleaner to use the mixpanel-browser library here like we do in dev console: https://github.com/ethereum-optimism/ecosystem-private/blob/main/apps/dev-console-interface/src/event-tracking/mixpanel.tsx#L10

packages/demo/frontend/src/utils/analytics.ts
```diff
@@ -0,0 +1,49 @@
...
+  interface Window {
+    mixpanel?: {
+      init: (token: string, config?: Record<string, unknown>) => void
+      track: (event: string, properties?: Record<string, unknown>) => void
+      identify: (userId: string) => void
+      people: { set: (properties: Record<string, unknown>) => void }
+      reset: () => void
+    }
+  }
+}
```
tremarkley: this can be deleted. see comments below

tremarkley (approved):
feel free to merge once you resolve my comments!

packages/demo/frontend/src/utils/analytics.ts
its-everdred: i did [originally](https://app.circleci.com/pipelines/github/ethereum-optimism/actions/458/workflows/13c40280-ae7e-405c-b6ea-b57283148ea0/jobs/613) and it blew through our CI memory: `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`

its-everdred: I can play with it though if we think the sdk is better

PR #260, opened 2025-12-17
Upgrade Morpho Demo Market

its-everdred:
## Problem

The demo morpho market is past the max limit and will no longer accept deposits:

## Solution

- [x] Add contracts to deploy a new DEMO_USDC token
- [x] Add contract to deploy a new DEMO_OP collateral token 
- [x] Set up a Morpho v1.1 market with these assets and a static price oracle
- [x] Create a script that deploys both tokens and sets up a market generating yield
- [x] Replace old DEMO_USDC with new asset market in the demo
- [x] Fix balance react state bug
- [x] Pull morpho APY from onchain instead of hardcoding demo APY
- [x] Fix env var name typo
- [x] Introduce `MorphoContractsRegistry` pattern for chain-specific protocol addresses
- [x] Add `actions.optimism.io` to CORS list 
- [ ] set `BASE_SEPOLIA_BUNDLER_URL` in prod


packages/demo/contracts/script/DeployMorphoMarket.s.sol
```diff
@@ -0,0 +1,80 @@
...
+        usdc.approve(vault, INITIAL_USDC_SUPPLY);
+        IMetaMorpho(vault).deposit(INITIAL_USDC_SUPPLY, msg.sender);
+
+        // Mint OP and supply as collateral
+        op.mint(msg.sender, COLLATERAL_AMOUNT);
+        op.approve(MORPHO, COLLATERAL_AMOUNT);
+        IMorpho(MORPHO).supplyCollateral(marketParams, COLLATERAL_AMOUNT, msg.sender, "");
+
+        // Borrow USDC to generate yield (99.9% utilization)
+        IMorpho(MORPHO).borrow(marketParams, BORROW_AMOUNT, 0, msg.sender, msg.sender);
```
tremarkley, line 76: whats the reasoning behind borrowing at a 99.9% utilization? doesnt that bring the position closer to being liquidated or will the positon never be liquidated because the demo will only be depositing and never borrowing from the market?

packages/demo/contracts/script/DeployMorphoMarket.s.sol
```diff
@@ -0,0 +1,80 @@
...
+import {IMorpho, IMetaMorpho, IMetaMorphoFactory, MarketParams} from "../src/interfaces/IMorpho.sol";
+
+/// @title DeployMorphoMarket
+/// @notice Deploys a complete Morpho lending market with yield generation in a single transaction.
+///         Uses MetaMorpho V1.1 factory which allows 0 timelock for instant cap acceptance.
+contract DeployMorphoMarket is Script {
+    address constant MORPHO = 0xBBBBBbbBBb9cC5e90e3b3Af64bdAF62C37EEFFCb;
+    address constant METAMORPHO_FACTORY_V1_1 = 0x2c3FE6D71F8d54B063411Abb446B49f13725F784;
+    address constant IRM = 0x46415998764C29aB2a25CbeA6254146D50D22687;
+    uint256 constant LLTV = 945000000000000000;
```
tremarkley, line 17: so if the LLTV drops below 94.5% what will happen to the market and subesquently the usability of the demo? what was the intention behind setting this at 94.5%

packages/demo/frontend/src/components/earn/Action.tsx
```diff
@@ -61,9 +61,23 @@ export function Action({
     marketId?.address.toLowerCase() ===
       '0x4200000000000000000000000000000000000006'
 
+  // Max withdrawal for illiquid Aave market
+  const AAVE_MAX_WITHDRAW = 0.0001
+
+  // Determine display precision based on asset type
+  const isWethAsset = assetSymbol === 'WETH' || assetSymbol.includes('ETH')
```
tremarkley: wait, so WETH is both ETH and WETH?

packages/demo/frontend/src/components/earn/Action.tsx
```diff
@@ -61,9 +61,23 @@ export function Action({
...
   const handleMaxClick = () => {
-    const maxAmount = mode === 'lend' ? assetBalance : depositedAmount || '0'
-    const rounded = parseFloat(maxAmount).toFixed(2)
+    let maxAmount = mode === 'lend' ? assetBalance : depositedAmount || '0'
+
+    // For illiquid Aave market in withdraw mode, cap at max withdrawal limit
+    if (isIlliquidAaveMarket && mode === 'withdraw') {
+      const deposited = parseFloat(depositedAmount || '0')
+      maxAmount = Math.min(deposited, AAVE_MAX_WITHDRAW).toString()
+    }
```
tremarkley, line 92: do we expect this to work even when the market is illiquid? or will it most likely fail?

packages/demo/frontend/src/components/earn/Action.tsx
```diff
@@ -446,49 +460,76 @@ export function Action({
             disabled={(() => {
               const maxAmount =
                 mode === 'lend' ? assetBalance : depositedAmount || '0'
+              const amountValue = parseFloat(amount)
+
+              // For illiquid Aave market, limit withdrawals
+              const exceedsAaveLimit =
+                isIlliquidAaveMarket &&
+                mode === 'withdraw' &&
+                amountValue > AAVE_MAX_WITHDRAW
```
tremarkley: nit: this could be stored on the component since you use it in multiple places

packages/demo/frontend/src/components/earn/Action.tsx
```diff
@@ -446,49 +460,76 @@ export function Action({
...
+                const exceedsAaveLimit =
+                  isIlliquidAaveMarket &&
+                  mode === 'withdraw' &&
+                  amountValue > AAVE_MAX_WITHDRAW
+                const isDisabled =
+                  isLoading ||
+                  !amount ||
+                  amountValue <= 0 ||
+                  amountValue > parseFloat(maxAmount) ||
+                  exceedsAaveLimit
```
tremarkley: nit: this is also used in multiple places and could be stored in a variable at top-level of the component.

packages/demo/frontend/src/utils/balanceMatching.ts
```diff
@@ -66,21 +66,28 @@ export function matchAssetBalance({
     )
   }
 
+  const isWeth =
+    selectedAssetSymbol === 'WETH' || selectedAssetSymbol.includes('ETH')
```
tremarkley: similar question here as above, are we considering both ETH and WETH to be WETH?

tremarkley (approved):
nice, exciting to see that we can calculate the apys onchain

packages/demo/contracts/script/DeployMorphoMarket.s.sol
its-everdred: LLTV is the per-position liquidation threshold that doesn't change. 94.5% was the example value from [morpho docs](https://docs.morpho.org/tools/offchain/sdks/blue-sdk/#basic-usage). LTV is the current, dynamic ratio, and below 94.5% is the healthy state. Our script set the market as waaay over-collateralized 100:1. So the market will accrue interest fees for many years/decades growing from 1% -> 94.5%  before being in liquidation range.

packages/demo/contracts/script/DeployMorphoMarket.s.sol
its-everdred: High utilization generates high yield for depositors so the demo number go up fast. The position shouldn't be liquidated because the oracle returns a fixed 1:1 price so collateral value never drops relative to debt and because we're way [over collateralized](https://github.com/ethereum-optimism/actions/pull/260#discussion_r2629080494)

packages/demo/frontend/src/components/earn/Action.tsx
its-everdred: From the few times I've tested it randomly over the last couple weeks, I believe withdrawing 0.0001 at a time has always succeeded. So now we just prevent the user from trying to withdraw more and show them this:  
image

packages/demo/frontend/src/components/earn/Action.tsx
its-everdred: This was needlessly broad, updated it to only `WETH`.

packages/demo/frontend/src/utils/balanceMatching.ts
its-everdred: Made this only WETH as well.

packages/demo/frontend/src/components/earn/Action.tsx
its-everdred: Fixed

packages/demo/frontend/src/components/earn/Action.tsx
its-everdred: fixed

PR #261, opened 2025-12-18
Small demo fixes

its-everdred:
- [x] Add Aave ETH market low liquidity message on tx fail
- [x] Update Mint USDC -> Mint Asset
- [x] State for faucet max per day

image

### DIFF
--- a/packages/demo/frontend/src/components/earn/Action.tsx
+++ b/packages/demo/frontend/src/components/earn/Action.tsx
@@ -53,6 +53,8 @@ export function Action({
   const [blockExplorerUrl, setBlockExplorerUrl] = useState<string | undefined>(
     undefined,
   )
+  const [showFaucetInfo, setShowFaucetInfo] = useState(false)
+  const [hasUsedWethFaucet, setHasUsedWethFaucet] = useState(false)
 
   // Check if this is the illiquid Aave OP Sepolia ETH market
   const isIlliquidAaveMarket =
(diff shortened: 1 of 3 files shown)


packages/demo/frontend/src/components/earn/Action.tsx
```diff
@@ -188,30 +192,72 @@ export function Action({
...
+                  }
+                  onMouseLeave={() => setShowFaucetInfo(false)}
                 >
-                  Get {displaySymbol}
-                </button>
+                  <button
+                    onClick={() => {
+                      trackEvent('mint_asset', { asset: assetSymbol })
+                      if (isWethAsset) {
+                        setHasUsedWethFaucet(true)
```
tremarkley, line 206: is there also a balance check somewhere that locks down access to the faucet? will this button be re-enabled when the page is refreshed?

tremarkley (approved):

packages/demo/frontend/src/components/earn/Action.tsx
its-everdred: Refreshing will re-enable, but clicking will turn it grey again. it will actually attempt the tx, but fail at the contract level with `Faucet: auth cannot be used yet because timeout has not elapsed`.

This is the quickest fix to give the user some idea why it's erroring without having the frontend check the chain timeout or storing timeout state.

PR #284, opened 2026-01-29
SwapProvider and Uniswap support

its-everdred:
## Problem

Actions doesn't offer the ability to swap between tokens.

## Solution

- [x] Add a `SwapProvider` abstraction
- [x] Add a `UniswapSwapProvider` implementation supporting Uni v4
- [x] Add swap support to the demo backend
- [x] Add swap support to the demo frontend
- [x] Redesign the frontend to allow many actions
- [x] Move balance to the nav, add mobile nav.
- [x] Add a new Activity section that summarizes the Logs
- [x] Add uniswap market deployment script used to set up demo markets


docs/specs/swap-provider-summary.md
```diff
@@ -0,0 +1,249 @@
...
+  recipient?: Address
+}
+```
+
+**Returns:** `SwapReceipt`
+
+```typescript
+interface SwapReceipt {
+  receipt: TransactionReceipt    // Blockchain transaction receipt
+  amountIn: bigint               // Actual input amount (wei)
```
Ponx: The amountIn and amountOut should mirror the format of the request amountIn and amountOut - that is, be human-readable. This way a developer doesn't have to convert these for display. (Or if the wallet is converting for display, then there should be more consistency because it would convert up to wei on the way out and convert down to human readable on the way in.)

Or, includes separate fields (suggest with different names) to show the wei amountIn and amountOut.

docs/specs/swap-provider-summary.md
```diff
@@ -0,0 +1,249 @@
...
+**Parameters:**
+
+```typescript
+interface SwapPriceParams {
+  /** Token to get price for */
+  assetIn: Asset
+  /** Token to price against. Defaults to USDC. */
+  assetOut?: Asset
+  /** Input amount. Defaults to 1 unit. */
+  amountIn?: number
```
Ponx: Nit: clarify only one of amountIn or amountOut can be specified, and that these should be the human-readable values.

Ponx: Nit: clarify these are human-readable (vs wei)

docs/specs/swap-provider-summary.md
```diff
@@ -0,0 +1,249 @@
...
+**Parameters:**
+
+```typescript
+interface SwapExecuteParams {
+  /** Amount of input token (human-readable). Required if amountOut not provided. */
+  amountIn?: number
+  /** Amount of output token for exact output swaps. Required if amountIn not provided. */
+  amountOut?: number
+  /** Token to sell */
+  assetIn: Asset
```
Ponx: For clarity, is Asset just a string, or an object with additional parameters?

docs/specs/swap-provider-summary.md
```diff
@@ -0,0 +1,249 @@
...
+}
+```
+
+**Returns:** `SwapPrice`
+
+```typescript
+interface SwapPrice {
+  price: string                  // Exchange rate (e.g., "3245.50")
+  priceInverse: string           // Inverse rate
+  amountIn: bigint               // Input amount (wei)
```
Ponx: amountIn should also have a formatted option, in case the request is for an exact out (amountOut is populated).

docs/specs/swap-provider-summary.md
its-everdred: Agreed! good call out

docs/specs/swap-provider-summary.md
its-everdred: It's an object with additional params, including the asset addresses on their chosen supported chains. The user imports or defines the assets they want to support and include them in their ActionsConfig. [example](https://actions.optimism.io/#assets)

packages/sdk/src/lend/providers/morpho/sdk.ts
```diff
@@ -345,7 +345,7 @@ export async function getVault(params: GetVaultParams): Promise<LendMarket> {
         params.marketId.address,
       ).catch((error) => {
         console.error('Failed to fetch rewards data:', error)
-        return { usdc: 0, morpho: 0, other: 0, totalRewards: 0 }
+        return { eth: 0, weth: 0, usdc: 0, usdc_demo: 0, op_demo: 0, morpho: 0, other: 0, totalRewards: 0 }
```
its-everdred: This is a bad design [known issue](https://github.com/ethereum-optimism/actions/issues/293) that shouldn't be hard coded.

packages/demo/backend/src/config/actions.ts
```diff
@@ -37,8 +37,16 @@ export function createActionsConfig(): NodeActionsConfig<'privy'> {
...
+      uniswap: {
+        defaultSlippage: 0.005, // 0.5%
+        marketAllowlist: [
+          { assets: [USDC_DEMO, OP_DEMO], fee: 100, tickSpacing: 2 },
+        ],
+      },
+    },
     assets: {
-      allow: [USDC_DEMO, ETH],
+      allow: [USDC_DEMO, OP_DEMO, ETH],
```
its-everdred, line 49: ℹ️  This is essentially the only backend code snippet devs will need to add uniswap support to their app.

packages/demo/backend/package.json
```diff
@@ -28,13 +28,14 @@
...
-    "lint:fix": "eslint \"**/*.{ts,tsx}\" --fix && prettier \"**/*.{ts,tsx}\" --write --loglevel=warn",
+    "lint:fix": "eslint \"**/*.{ts,tsx}\" --fix && prettier \"**/*.{ts,tsx}\" --write --log-level=warn",
     "dev": "tsx src/index.ts",
     "dev:local": "LOCAL_DEV=true pnpm dev",
     "start": "node dist/index.js",
     "test": "vitest --run",
     "typecheck": "tsc --noEmit",
-    "attribution": "tsx scripts/attribution.ts"
+    "attribution": "tsx scripts/attribution.ts",
+    "deploy:uniswap": "tsx scripts/deployUniswapMarket.ts"
```
its-everdred, line 38: This will likely be replaced with an Actions CLI command in the future

packages/demo/contracts/script/DeployUniswapMarket.s.sol
```diff
@@ -0,0 +1,128 @@
...
+            // 1 USDC = 5.5556 OP → raw price = 5.5556e12
+            // sqrtPriceX96 = sqrt(5.5556e12) * 2^96
+            return 186742601293858871787747742452809728;
+        } else {
+            // token0 = OP, token1 = USDC
+            // 1 OP = 0.18 USDC → raw price = 1.8e-13
+            // sqrtPriceX96 = sqrt(1.8e-13) * 2^96
+            return 33613662584877347604918;
+        }
+    }
```
its-everdred, line 127: This `sqrtPrice` calculation is the Uniswap v4 complexity you've likely heard of.

packages/demo/contracts/package.json
```diff
@@ -11,10 +11,14 @@
...
     "impersonate:fund:faucet": "pnpm impersonate:whale && forge script script/ImpersonateFund.s.sol --rpc-url http://127.0.0.1:9545 --broadcast --unlocked 0x5752e57DcfA070e3822d69498185B706c293C792 --sender 0x5752e57DcfA070e3822d69498185B706c293C792",
     "deploy:impersonate:fund:faucet": "FUND_FAUCET_ETH=true pnpm deploy:faucet:supersim && pnpm impersonate:fund:faucet",
+    "deploy:demo": "bash script/deploy-demo.sh",
     "test": "forge install && forge test",
     "lint": "solhint {src,script,test}/**/*.sol && forge fmt --check src test scripts",
     "lint:fix": "solhint --fix --quiet --noPrompt {src,script,test}/**/*.sol && forge fmt src test script"
   },
+  "nx": {
+    "includedScripts": []
+  },
```
its-everdred, line 21: ℹ️ makes the Foundry contracts package not required from workspace-wide pnpm -r commands since forge isn't needed in all CI sub-tasks. I would leave this as a comment, but json 😬

packages/demo/frontend/src/constants/routes.ts
```diff
@@ -1,7 +1,6 @@
 export const ROUTES = {
   HOME: '/',
   DEMO: '/demo',
-  DOCS: '/docs',
```
its-everdred, line 4: These docs were very old

packages/demo/backend/src/controllers/swap.ts
```diff
@@ -0,0 +1,145 @@
...
+import type { Context } from 'hono'
+import type { Address } from 'viem'
+import { z } from 'zod'
+
+import { serializeBigInt } from '@/utils/serializers.js'
+
+import { errorResponse, requireAuth } from '../helpers/errors.js'
+import { validateRequest } from '../helpers/validation.js'
+import * as swapService from '../services/swap.js'
+
```
tremarkley, line 13: nit: use `@` imports. this applies to all the imports throughout this PR

packages/demo/backend/package.json
```diff
@@ -27,14 +27,14 @@
   "scripts": {
     "build": "pnpm clean && tsc && resolve-tspaths",
     "clean": "rm -rf dist tsconfig.tsbuildinfo",
-    "lint": "eslint \"**/*.{ts,tsx}\" && prettier --check \"**/*.{ts,tsx}\"",
-    "lint:fix": "eslint \"**/*.{ts,tsx}\" --fix && prettier \"**/*.{ts,tsx}\" --write --loglevel=warn",
+    "lint": "eslint \"**/*.{ts,tsx}\" --fix && prettier \"**/*.{ts,tsx}\" --write --log-level=warn",
```
tremarkley: hmm for ci purposes we dont want to include `fix` in the `lint` command, right? otherwise, lint could be failing without us knowing.

packages/demo/contracts/package.json
tremarkley: does having the foundry contracts included in workspace-wide pnpm commands cause ci issues? does this mean separate ci checks should be run specifically for the contracts package?

packages/demo/frontend/src/components/earn/Action.tsx
tremarkley, line 1: nit: there are multiple components inside this file (`ModeToggle`, `AmountLabel`, etc) consider moving them into their own files to make this more readable / maintainable

packages/sdk/src/swap/core/SwapProvider.ts
```diff
@@ -0,0 +1,252 @@
...
+  /**
+   * Execute a token swap
+   * @param params - Swap parameters including assets, amounts, and chain
+   * @returns Swap transaction data ready for execution
+   */
+  async execute(
+    params: SwapExecuteParams & {
+      walletAddress: Address
+      chainId: SupportedChainId
+    },
```
tremarkley: how come `walletAddress` and `chainId` are not included in `SwapExecuteParams`?

packages/sdk/src/swap/core/SwapProvider.ts
```diff
@@ -0,0 +1,252 @@
...
+  validateChainSupported,
+  validateNotBothAmounts,
+  validateNotSameAsset,
+  validateNotZeroAddress,
+  validateSlippage,
+} from '@/utils/validation.js'
+
+const DEFAULT_SLIPPAGE = 0.005
+const DEFAULT_DEADLINE_OFFSET = 60
+const MAX_SLIPPAGE = 0.5
```
tremarkley: nit: should `MAX_SLIPPAGE` be configurable?

packages/sdk/src/swap/core/SwapProvider.ts
```diff
@@ -0,0 +1,252 @@
...
+        assetOut,
+        chainId,
+        marketAllowlist,
+      )
+      if (!isAllowed) {
+        throw new Error(
+          `Pair ${assetIn.metadata.symbol}/${assetOut.metadata.symbol} is not in the allowlist for chain ${chainId}`,
+        )
+      }
+    }
```
tremarkley, line 145: nit: if `!marketAllowlist.length` should this throw an error?

packages/sdk/src/swap/core/SwapProvider.ts
```diff
@@ -0,0 +1,252 @@
...
+      return this.configContainsPair(
+        addressIn,
+        addressOut,
+        chainId,
+        config.assets,
+      )
+    })
+  }
+
+  private configContainsPair(
```
tremarkley: nit: the naming of this is a little confusing since it isnt actually checking `this.config` for the pair

packages/sdk/src/swap/namespaces/ActionsSwapNamespace.ts
```diff
@@ -0,0 +1,7 @@
+import { BaseSwapNamespace } from './BaseSwapNamespace.js'
```
tremarkley: nit: use @ import

packages/sdk/src/swap/namespaces/BaseSwapNamespace.ts
```diff
@@ -0,0 +1,76 @@
...
+      for (const chainId of provider.supportedChainIds()) {
+        chainIds.add(chainId)
+      }
+    }
+    return Array.from(chainIds)
+  }
+
+  protected getAllProviders(): Array<SwapProvider<SwapProviderConfig>> {
+    return Object.values(this.providers).filter(
+      (p): p is SwapProvider<SwapProviderConfig> => p !== undefined,
```
tremarkley, line 65: does this mean that `this.providers` could contain `undefined` entries?

packages/sdk/src/swap/namespaces/WalletSwapNamespace.ts
```diff
@@ -0,0 +1,83 @@
+import type { SupportedChainId } from '@/constants/supportedChains.js'
+import type {
+  SwapExecuteParams,
+  SwapProviders,
+  SwapReceipt,
+  SwapTransaction,
+} from '@/types/swap/index.js'
+import type { Wallet } from '@/wallet/core/wallets/abstract/Wallet.js'
+
+import { BaseSwapNamespace } from './BaseSwapNamespace.js'
```
tremarkley: nit: use @ import

packages/sdk/src/swap/providers/uniswap/encoding.ts
```diff
@@ -0,0 +1,400 @@
...
+ */
+function resolvePoolParams(
+  assetIn: Asset,
+  assetOut: Asset,
+  chainId: SupportedChainId,
+  fee: number,
+  tickSpacing: number,
+): ResolvedPoolParams {
+  const tokenIn = isNativeAsset(assetIn)
+    ? getAssetAddress(WETH, chainId)
```
tremarkley: why does native asset use the `WETH` address? isn't the native asset `ETH`?

packages/sdk/src/swap/providers/uniswap/UniswapSwapProvider.ts
```diff
@@ -0,0 +1,364 @@
...
+  // ─────────────────────────────────────────────────────────────────────────────
+
+  /**
+   * Check ERC20 and Permit2 allowances, returning approval txs if needed.
+   * Skipped entirely for native ETH swaps.
+   * @param params - Resolved swap params (wallet address, asset info)
+   * @param requiredAmount - Amount in wei that must be approved
+   * @returns Token and Permit2 approval transactions (undefined if not needed)
+   */
+  private async buildApprovals(
```
tremarkley: is this unique to uniswap or could it also be shared with other swap providers? the only part that looks uniswap-specific is the call to `getUniswapAddresses`

packages/sdk/package.json
```diff
@@ -41,8 +41,7 @@
     "build": "tsc && resolve-tspaths",
     "clean": "rm -rf dist types tsconfig.tsbuildinfo",
     "docs": "typedoc",
-    "lint": "eslint \"**/*.{ts,tsx}\" && prettier --check \"**/*.{ts,tsx}\"",
-    "lint:fix": "eslint \"**/*.{ts,tsx}\" --fix && prettier \"**/*.{ts,tsx}\" --write --loglevel=warn",
+    "lint": "eslint \"**/*.{ts,tsx}\" --fix && prettier \"**/*.{ts,tsx}\" --write --log-level=warn",
```
tremarkley: same comment here as above, I think that including `--fix` in the lint command has potential to swallow lint errors in ci

tremarkley (approved):
just adding a reminder in case you havent already done this - it would be good to test that the demo app works across all the wallet providers

packages/sdk/src/swap/core/SwapProvider.ts
its-everdred: good call

packages/sdk/src/swap/providers/uniswap/encoding.ts
its-everdred: oh wow, great catch! I set this because v3 always uses WETH under the hood. I forgot v4 supports native ETH directly, so this would have thrown an error. updated

packages/demo/backend/package.json
its-everdred: oh yeah good call, i was being lazy trying to simplify but forgot we do need both

packages/sdk/src/swap/core/SwapProvider.ts
its-everdred: It was already configurable by the caller, but I just added the ability for the `ActionsConfig` to override it too

packages/sdk/src/swap/core/SwapProvider.ts
its-everdred: good idea

packages/sdk/src/swap/providers/uniswap/UniswapSwapProvider.ts
its-everdred: good call, pulled out and shared

packages/sdk/src/swap/namespaces/BaseSwapNamespace.ts
its-everdred: Yes, provider keys are optional since not all providers need to be configured. We enforce at least one provider via `RequireAtLeastOne<SwapConfig>` at the type level. Added a comment clarifying this.

its-everdred:
> does having the foundry contracts included in workspace-wide pnpm commands cause ci issues? does this mean separate ci checks should be run specifically for the contracts package?

pnpm -r typecheck skips packages that don't have a typecheck script. The `"nx": { "includedScripts": [] }` tells Nx to ignore it during nx run-many commands. Between these two, the contracts package is excluded from workspace-level CI checks.

packages/sdk/src/swap/core/SwapProvider.ts
its-everdred: Address is passed separately by `WalletSwapNamespace`, but this is confusing, so I made separate types that include all the params for each function.

packages/demo/backend/src/controllers/swap.ts
its-everdred: added this as a lint warning so I don't keep forgetting

tremarkley:
> pnpm -r typecheck skips packages that don't have a typecheck script. The "nx": { "includedScripts": [] } tells Nx to ignore it during nx run-many commands. Between these two, the contracts package is excluded from workspace-level CI checks.

@its-everdred dont we still want the contracts package included in workspace-level CI checks? if you remove these, would it be possible for someone to push up a change that breaks the build of the contracts package?

its-everdred:
> dont we still want the contracts package included in workspace-level CI checks? if you remove these, would it be possible for someone to push up a change that breaks the build of the contracts package?

@tremarkley that's true. I just broke contracts into a separate CI step for build + tests.

PR #288, opened 2026-02-07
Borrow spec

its-everdred:
## Problem

Actions SDK offers lending but not borrowing

## Solution

Adds a few spec files to scope out borrow:

- [ ] Summary spec - defining interfaces
- [ ] Detailed spec - defining implementation plan
- [ ] Aave spec - defining adding Aave borrowing support


jefr90:
Thanks. Closing: borrow work is now in the active stack (#458 / #465 / #466).

PR #292, opened 2026-02-10
Tree-shakeability

its-everdred:
# Problem
The SDK forces all provider dependencies into every consumer's bundle, regardless of whether they even want to use all sub-dependencies.

# Solution

Unclear. For now I discuss some options in a spec.

A Solution test should look something like: 
1. Create a new node package
2. Add `actions-sdk`
4. Create an `ActionsConfig` and only use/install `morpho` and `turnkey`.

**Before**: 
Dependency install size: X kb 
Runtime size: Y kb 

**After**: 
Dependency install size: A kb 
Runtime size: B kb


jefr90:
Thanks. Closing as stale (no activity since March). Reopen when there's bandwidth.

PR #299, opened 2026-03-04
Backend preview builds

its-everdred:
## Problem

The Uniswap branch introduces new demo backend endpoints, but we have no way to test in preview build because we don't deploy preview backend builds.

## Solution

Add ephemeral backend deploy previews by wrapping the Hono app as a Netlify Function so PR preview URLs serve both frontend and backend from the same origin.

- [ ] Add staging backend env vars to netlify


jefr90:
Thanks. Closing: CircleCI failing since March, stale since then. Reopen or file fresh if still relevant.

PR #326, opened 2026-03-23
feat(sdk): add EIP-55 address validation for hardcoded and config addresses

### LINKED ISSUE #323: Add EIP-55 checksum validation for hardcoded contract addresses
## Problem

All protocol contract addresses in the SDK are hardcoded as string literals without checksum validation. This means:

1. **Typos and truncation go undetected** — A malformed address (e.g., 41 chars instead of 42) compiles fine and only fails at runtime when a transaction is sent
2. **No EIP-55 mixed-case checksum enforcement** — Copy/paste errors that change casing are not caught

This was a real issue in #319, where the Ink Aave V3 Pool address was missing a character (`0x2816cf15F6d2220E...` instead of `0x2816cf15F6d2A220E...`) and passed type-checking.

## Affected Files

- `packages/sdk/src/lend/providers/aave/addresses.ts` — 6 chain entries × 4 addresses each
- `packages/sdk/src/lend/providers/morpho/contracts.ts` — 8 chain entries × 2 addresses each
- `packages/sdk/src/swap/providers/uniswap/addresses.ts` — 8 chain entries × 5 addresses each
- `packages/sdk/src/constants/assets.ts` — ~12 token definitions with per-chain addresses
- `packages/sdk/src/constants/contracts.ts` — Permit2 address

## Possible Solution

Rather than wrapping every individual address string in `getAddress()`, we might validate entire address maps at the point of definition using a shared utility:

```typescript
// packages/sdk/src/utils/validateAddresses.ts
import { getAddress, type Address } from 'viem'
import type { SupportedChainId } from '@/constants/supportedChains.js'

/**
 * Validates all addresses in a per-chain address map.
 * Throws at module load time if any address is malformed.
 */
export function validateAddressMap<T extends Record<string, Address>>(
  map: Partial<Record<SupportedChainId, T>>,
): Partial<Record<SupportedChainId, T>> {
  for (const [chainId, addresses] of Object.entries(map)) {
    for (const [key, addr] of Object.entries(addresses as Record<string, Address>)) {
      try {
        getAddress(addr)
      } catch {
        throw new Error(
          `Invalid address for ${key} on chain ${chainId}: ${addr}`,
        )
      }
    }
  }
  return map
}

/**
 * Validates all addresses in an Asset.address map.
 * Skips 'native' sentinel values.
 */
export function validateAssetAddresses(
  map: Partial<Record<SupportedChainId, Address | 'native'>>,
): Partial<Record<SupportedChainId, Address | 'native'>> {
  for (const [chainId, addr] of Object.entries(map)) {
    if (addr !== 'native') {
      try {
        getAddress(addr)
      } catch {
        throw new Error(`Invalid asset address on chain ${chainId}: ${addr}`)
      }
    }
  }
  return map
}
```

Then each protocol file wraps its map with a single call — no changes to individual address strings:

```typescript
// aave/addresses.ts
const AAVE_ADDRESSES = validateAddressMap<AaveAddresses>({ ... })

// morpho/contracts.ts
export const MORPHO_CONTRACTS: MorphoContractsRegistry = validateAddressMap<MorphoContracts>({ ... })

// uniswap/addresses.ts
const UNISWAP_ADDRESSES = validateAddressMap<UniswapAddresses>({ ... })

// assets.ts
export const USDC: Asset = {
  address: validateAssetAddresses({ ... }),
  ...
}
```

**Total changes:** ~1 new utility file + ~1 line per protocol/asset definition. No touching individual address strings.

## Scope

- SDK-internal hardcoded addresses only
- Developer-supplied address validation is tracked separately in #324

### LINKED ISSUE #324: Validate developer-supplied addresses at SDK initialization
## Problem

Addresses passed through `ActionsConfig` by SDK consumers (market allowlists, custom asset definitions) are never validated. A malformed or truncated address silently fails — either never matching anything in comparisons, or worse, being sent on-chain and wasting gas.

Currently the SDK does case-insensitive comparison (`.toLowerCase()`) at every address comparison site, which handles casing differences. But it does **not** check that addresses are valid 42-character hex strings.

## Affected Inputs

Developer-supplied addresses enter through:

- **`LendProviderConfig.marketAllowlist[].address`** — Morpho vault addresses, Aave reserve addresses
- **`LendProviderConfig.marketBlocklist[].address`** — Same shape
- **`SwapMarketConfig` addresses** — Swap market definitions
- **`Asset.address` entries** — Custom asset definitions with per-chain addresses
- **`LendMarketConfig.asset.address`** — Asset addresses nested inside market configs

These flow through `LendProvider.validateConfigSupported()`, `SwapProvider._validateSwapConfigSupported()`, and into on-chain calls via `getAssetAddress()`.

## Comparison Sites (already case-insensitive, but no format validation)

- `LendProvider.ts:236` — `allowedMarket.address.toLowerCase() === marketId.address.toLowerCase()`
- `BaseLendNamespace.ts:79` — same pattern
- `aave/sdk.ts:56` — same pattern
- `morpho/sdk.ts:303` — same pattern
- `SwapProvider.ts:292-296` — `.toLowerCase()` for swap address matching

## Proposed Solution

Add address validation at the SDK initialization boundary — when `ActionsConfig` is processed in the `Actions` constructor or a dedicated config validation step.

```typescript
import { getAddress } from 'viem'

function validateAddress(address: string): Address {
  try {
    return getAddress(address)
  } catch {
    throw new Error(`Invalid address: ${address}`)
  }
}
```

Apply to:
1. All `LendMarketConfig.address` entries in `marketAllowlist` / `marketBlocklist`
2. All `Asset.address` values in custom asset definitions
3. Any other developer-supplied addresses entering the SDK

This catches bad input early with a clear error message, rather than silently failing downstream.

## Scope

- Runtime validation only (developer-supplied input)
- Internal hardcoded address validation is tracked separately in #323
- No behavioral change for valid addresses — `getAddress()` normalizes to EIP-55 checksum format, and all comparisons already use `.toLowerCase()`

jefr90:
Closes #323, #324

## Summary
- Adds `packages/sdk/src/utils/validateAddresses.ts` with three exported functions:
  - `validateAddressMap` — validates `Partial<Record<number, Address | NamedAddresses>>` maps; collects all failures before throwing
  - `validateAssetAddresses` — same collect-all pattern for `Partial<Record<SupportedChainId, Address | 'native'>>`, skipping `'native'` entries
  - `validateConfigAddresses` — validates all developer-supplied addresses in an `ActionsConfig`; called in the `Actions` constructor before any provider is instantiated
- Validates hardcoded address maps via CI tests (#323): `validateAddresses.test.ts` imports and validates all internal address maps (`POOL_ADDRESSES_*`, `WETH_GATEWAY_ADDRESSES_*`, `MORPHO_CONTRACTS`, `UNISWAP_ADDRESSES`, all asset constants) — catching bad addresses at test time without running validation at every consumer app startup
- Calls `validateConfigAddresses(config)` in the `Actions` constructor for developer-supplied config (#324)
- `validateConfigAddresses` iterates providers generically via `Object.entries(config.lend ?? {})` / `Object.entries(config.swap ?? {})` — new lend or swap providers added to `LendConfig`/`SwapConfig` are automatically validated without touching this function
- Adds changeset for `@eth-optimism/actions-sdk`

## Error behaviour

**#323 — malformed hardcoded address (CI / test time)**

All internal address maps are validated by `validateAddresses.test.ts`. A bad address in e.g. `POOL_ADDRESSES_MAINNET` fails CI before it can ship, rather than erroring at consumer startup or silently producing a wrong transaction at runtime.

**#324 — malformed developer-supplied address (SDK initialisation)**

If an address passed via `ActionsConfig` is invalid, the `Actions` constructor throws before any provider is instantiated. All invalid addresses across all config sections are collected into a single error:

```
Error: Invalid addresses found:
  - lend.morpho.marketAllowlist[0].address: 0xabc (not a valid EVM address)
  - assets.allow[0].address[10]: 0xdeadbeef (not a valid EVM address)
```

## Test plan
- [x] All 364 unit tests pass
- [x] New CI tests validate all hardcoded address maps and asset constants
- [x] `validateAddressMap` / `validateAssetAddresses`: throw on malformed addresses, collect multiple failures, pass for valid and lowercase addresses, return original map reference
- [x] `validateConfigAddresses`: throws per section with descriptive messages, collects across sections, no-throw on valid config
- [x] `Actions` constructor: throws on invalid addresses in config, constructs successfully on valid config


its-applekid (approved):
Strong PR! A few observations:

**What I like:**
- Fail-fast at module load for hardcoded addresses → catches dev errors immediately
- Batch error collection → shows all problems at once rather than failing on first bad address
- Comprehensive coverage → wraps all address maps (Aave, Morpho, Uniswap, assets, config)
- Good test coverage → 18 tests including integration with Actions constructor
- Clear error messages → location + chain ID + invalid value

**Minor notes:**
- `validateAddressMap` generic works for both single Address and NamedAddresses record → nice type flexibility
- `validateAssetAddresses` correctly skips `'native'` entries
- Future dynamic config traversal idea (in PR description) makes sense if more providers are added

**Tested:**
- [x] All 357 existing tests pass
- [x] 18 new tests cover all validation functions
- [x] Constructor integration verified

Ready to merge. Nice work on the thorough testing and error UX!

its-everdred:
@jefr90 Great point about dynamic config traversal! This is actually relatively pressing since we're in the process of adding support for Velodrome. Ideally, I don't want new providers opening PRs to have to know about this detail, so it might be worth dealing with now while it's top of mind. As you mentioned, one solution could look something like this, but don't feel like you have to use it:

```
// Shared leaf helper — skips 'native'
function assetAddresses(asset: Asset, path: string) {
  return Object.entries(asset.address)
    .filter(([, addr]) => addr !== 'native')
    .map(([chainId, addr]) => ({ path: `${path}.address[${chainId}]`, value: addr as string }))
}

// Morpho + Aave — covers both market contract address and asset per-chain map
function lendProviderAddresses(config: LendProviderConfig, providerPath: string) {
  return ['marketAllowlist', 'marketBlocklist'].flatMap(key =>
    (config[key as keyof LendProviderConfig] as LendMarketConfig[] ?? []).flatMap((m, i) => [
      { path: `${providerPath}.${key}[${i}].address`, value: m.address },
      ...assetAddresses(m.asset, `${providerPath}.${key}[${i}].asset`),
    ])
  )
}

// Uniswap + Velodrome — VelodromeMarketConfig extends SwapMarketConfig, same assets[] shape
function swapProviderAddresses(config: SwapProviderConfig, providerPath: string) {
  return ['marketAllowlist', 'marketBlocklist'].flatMap(key =>
    (config[key as keyof SwapProviderConfig] as SwapMarketConfig[] ?? []).flatMap((m, i) =>
      m.assets.flatMap((asset, j) =>
        assetAddresses(asset, `${providerPath}.${key}[${i}].assets[${j}]`)
      )
    )
  )
}

function validateConfigAddresses(config: ActionsConfig): void {
  const addresses = [
    ...Object.entries(config.lend ?? {}).flatMap(([name, cfg]) =>
      lendProviderAddresses(cfg as LendProviderConfig, `lend.${name}`)
    ),
    ...Object.entries(config.swap ?? {}).flatMap(([name, cfg]) =>
      swapProviderAddresses(cfg as SwapProviderConfig, `swap.${name}`)
    ),
    ...(config.assets?.allow ?? []).flatMap((a, i) => assetAddresses(a, `assets.allow[${i}]`)),
    ...(config.assets?.block ?? []).flatMap((a, i) => assetAddresses(a, `assets.block[${i}]`)),
  ]
  // collect failures, throw once...
}
```

Wallet provider configs (`DefaultSmartWalletProvider`, hosted providers) don't contain EVM addresses, so no traversal needed there. And since this only runs once in the Actions constructor, the traversal cost is negligible.

its-everdred:
🤔 Okay, having thought about this more: In regards to the internal addresses (not the `ActionsConfig` ones), I think I sent us partially down the wrong path. I don't think we need every sdk consumer app to validate every token and every contract on every boot. Instead, we really only need to check that we aren't introducing bad addresses at CI time.  This could just be tests like:

```
// validateAddresses.test.ts (new section)
import { POOL_ADDRESSES_MAINNET, POOL_ADDRESSES_TESTNET } from '../lend/providers/aave/addresses'
import { MORPHO_CONTRACTS } from '../lend/providers/morpho/contracts'
import { UNISWAP_ADDRESSES } from '../swap/providers/uniswap/addresses'
import { ETH, WETH, USDC, MORPHO } from '../constants/assets'

it('all hardcoded address maps contain valid EVM addresses', () => {
  expect(() => validateAddressMap(POOL_ADDRESSES_MAINNET)).not.toThrow()
  expect(() => validateAddressMap(MORPHO_CONTRACTS)).not.toThrow()
  // etc.
})
```

We'll still want to validate addresses passed in by consumer apps via the `ActionsConfig` though.

jefr90:
Applied changes for both dynamic config traversal and moving internal address check to CI @its-everdred please let me know if it looks good

its-everdred (approved):
🙌 Looks great, thanks for making those changes!

PR #329, opened 2026-03-25
Optimization: Reduce bundle size

its-everdred:
## Problem

Frontend builds produce warnings:

```
(!) Some chunks are larger than 500 kB after minification.
```

The largest offenders exceed 2MB:

```
actionsApi-Bo6U1spp.js                 2,022.80 kB
privy-provider-BG8GtKO6-Kdn9sR6B.js   1,385.64 kB
DynamicProvider-Bv6AxASU.js            1,290.75 kB
```

The root cause is that `createActions` statically imports all wallet provider SDKs (Dynamic, Privy, Turnkey) via the `ReactHostedWalletProviderRegistry`, regardless of which provider the developer configured.

### 1. `sideEffects: false` on SDK package.json

Tells bundlers (Vite/Rollup/esbuild) that the SDK's modules have no side effects, enabling them to drop unused barrel re-exports. Reduced the `actionsApi` chunk from **2,022 kB → 110 kB**.

### 2. Lazy wallet provider initialization with dynamic imports

The wallet provider registries previously used static imports for all wallet providers at the top of the file. Now:

- Each registry factory uses `await import()` to load provider code on demand
- `WalletNamespace` lazily initializes the wallet provider on first method call (all wallet methods are already async)
- `createActions` **remains synchronous** — no consumer API change
- `ActionsConfig` shape is unchanged

The only interface change is that `actions.wallet.hostedWalletProvider` and `actions.wallet.smartWalletProvider` on `WalletNamespace` are now async methods instead of sync getters. These are advanced escape hatches not used in typical integrations.

**Result:** `EarnWithFrontendWallet` chunk dropped from **2,189 kB → 1,550 kB** (~30% reduction). Wallet providers now code-split into their own chunks (e.g. `DynamicHostedWalletProvider` at 611 kB).

Further tree-shaking research is tracked in [#292](https://github.com/ethereum-optimism/actions/pull/292).

## Test plan

- [x] All 364 SDK tests pass
- [x] Full NX build passes (SDK + frontend + backend)
- [x] Typecheck clean
- [x] Lint clean

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -11,6 +11,7 @@
   },
   "type": "module",
   "sideEffects": false,
+  "sideEffects": false,
   "version": "0.3.0",
   "description": "TypeScript SDK for Actions",
   "main": "dist/index.js",
--- a/packages/sdk/src/actions.ts
+++ b/packages/sdk/src/actions.ts
@@ -57,8 +57,6 @@ export class Actions<
     uniswap?: SwapProvider<SwapProviderConfig>
   } = {}
   private _assetsConfig?: AssetsConfig
-  private hostedWalletProvider!: THostedWalletProvidersSchema['providerInstances'][THostedWalletProviderType]
-  private smartWalletProvider!: SmartWalletProvider
   private hostedWalletProviderRegistry: HostedWalletProviderRegistry<
     THostedWalletProvidersSchema['providerInstances'],
     THostedWalletProvidersSchema['providerConfigs'],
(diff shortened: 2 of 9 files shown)


its-applekid:
## Review: Tree-shaking optimization

**Changes:**
1. `sideEffects: false` in package.json — Enables dead code elimination (2MB → 110kB chunk)
2. Lazy wallet provider init — Dynamic imports split providers into separate chunks (~30% reduction)
3. `WalletNamespace.hostedWalletProvider` / `smartWalletProvider` now async methods (advanced escape hatches only)

**Verified:**
- ✅ No global mutations, polyfills, or side-effect imports
- ✅ All 364 tests pass
- ✅ Typecheck + lint clean
- ✅ `createActions` stays sync — no consumer API changes

**Result:** Ready to merge. Low-risk, high-impact optimization following ecosystem best practices.

jefr90 (approved):
Do we want to commit .md's? If so, LGTM

PR #339, opened 2026-03-27
Replace `any` types with specific types across SDK and demo packages

### LINKED ISSUE #337: Replace all `any` types with specific types
## Problem

There are ~51 `@typescript-eslint/no-explicit-any` warnings across the SDK and demo backend. Using `any` bypasses type checking and can hide bugs.

## Scope

**SDK source files:**
- `lend/providers/aave/sdk.ts`
- `lend/providers/morpho/api.ts`
- `lend/providers/morpho/sdk.ts`
- `services/__mocks__/MockChainManager.ts`

**SDK test files:**
- `lend/core/__tests__/LendProvider.test.ts`
- `lend/providers/morpho/__tests__/MorphoLendProvider.test.ts`
- `services/tokenBalance.spec.ts`
- `swap/providers/uniswap/__tests__/sdk.test.ts`
- `utils/assets.test.ts`
- `wallet/react/wallets/hosted/dynamic/__tests__/DynamicWallet.spec.ts`
- `wallet/react/wallets/hosted/privy/__tests__/PrivyWallet.spec.ts`
- `wallet/core/wallets/smart/default/__tests__/DefaultSmartWallet.spec.ts`

**Demo backend:**
- `packages/demo/backend/src/helpers/validation.spec.ts`
- `packages/demo/backend/src/services/lend.spec.ts`
- `packages/demo/backend/src/services/swap.spec.ts`

## Guidelines

Follow existing patterns in the codebase:

- **External API responses** — use `unknown` and narrow with type guards or assertions, not `any`. See `utils/assets.ts` for examples of typed return values.
- **Mock objects in tests** — use `as unknown as SpecificType` instead of `as any`. See `swap/__mocks__/MockSwapProvider.ts` for the pattern:
  ```ts
  const publicClient = { readContract: vi.fn() } as unknown as PublicClient
  ```
- **Generic callbacks and parameters** — use `unknown` with type narrowing, or define a proper generic. Prefer `Record<string, unknown>` over `Record<string, any>`.
- **viem contract call returns** — cast to the expected tuple type. See `encoding/v2.ts`:
  ```ts
  return (amounts as bigint[])[1]
  ```
- **Error handling** — use `unknown` for catch parameters:
  ```ts
  catch (error: unknown) { ... }
  ```
- **JSON/API data** — define response interfaces rather than using `any`. See `types/swap/base.ts` and `types/lend/base.ts` for examples.

## Verification

After replacing all `any` types, `pnpm lint` should show 0 `no-explicit-any` warnings. Run `pnpm typecheck` to confirm no type errors were introduced.

daatsuka:
This change sweeps through 23 files in `packages/sdk`, `packages/demo/backend`, and `packages/demo/frontend` to eliminate every `@typescript-eslint/no-explicit-any` suppression. In the SDK source layer, the biggest wins are in `lend/providers/morpho/api.ts` and `lend/providers/morpho/sdk.ts`: `fetchRewards` now returns a properly typed `MorphoVaultApiResponse` instead of `any`, backed by four new interfaces (`MorphoRewardAsset`, `MorphoReward`, `MorphoMarketAllocation`, `MorphoVaultState`) that model the Morpho GraphQL schema. `calculateBaseApy`, `calculateApyBreakdown`, and `calculateRewardsBreakdown` in `sdk.ts` receive a structural `MorphoSdkVault` interface instead of `any`, and the inline `AccrualPosition` cast is gone — the reducer callback and `allocation.position` are now typed through `MorphoSdkAllocation`/`MorphoSdkPosition`. In `aave/sdk.ts`, `calculateApyBreakdown`'s `formattedReserve` parameter moves from `any` to `{ supplyAPY?: string }`. Over in `utils/test.ts`, the Anvil RPC calls (`anvil_impersonateAccount` / `anvil_stopImpersonatingAccount`) are handled via a structural cast on `publicClient.request` instead of casting the method string to `any`, and the redundant `as any` on `createWalletClient` is simply removed because viem's types already match.

In test files the pattern is consistent: every `as any` mock cast becomes `as unknown as <ConcreteType>`, where `<ConcreteType>` is derived from the function's actual return type — e.g. `ReturnType<typeof getActions>`, `Awaited<ReturnType<typeof fetchAccrualVault>>`, `Context` (from Hono), `PublicClient`, `BundlerClient`, `Asset`, `SupportedChainId`, or `Awaited<ReturnType<typeof toCoinbaseSmartAccount>>`. This two-step cast through `unknown` is the standard TypeScript escape hatch when building partial mocks: it satisfies the compiler while still documenting the intended type at the call site. The Morpho API test (`api.test.ts`) adds narrowing guards (`if (!vaultData) throw new Error('unreachable')`) before property access so the non-null assertions are explicit rather than silently swallowed by `any`. The `DynamicWallet.spec.ts` error-cause access switches from `(err as any).cause` to `(err as Error & { cause?: Error }).cause`, preserving type safety on the chained cause.

The design intent is to make every `any` removal a local, minimal change — introduce the narrowest structural interface that satisfies the call site, prefer `ReturnType`/`Awaited` utility types over hand-rolled duplicates, and avoid pulling in heavy external type packages. The Solidity file `DeployMorphoMarket.s.sol` only has a formatting adjustment that crept in from the linter pass. I ran `pnpm typecheck` and `pnpm lint` across the entire monorepo after the changes, confirming zero type errors and that the `no-explicit-any` warning count drops to zero in all touched files.

Closes https://github.com/ethereum-optimism/actions/issues/337

### DIFF
--- a/packages/demo/backend/src/helpers/validation.spec.ts
+++ b/packages/demo/backend/src/helpers/validation.spec.ts
@@ -1,3 +1,4 @@
+import type { Context } from 'hono'
 import { describe, expect, it, vi } from 'vitest'
 import { z } from 'zod'
 
@@ -40,7 +41,10 @@ describe('validateRequest', () => {
         false,
       )
 
-      const result = await validateRequest(mockContext as any, ParamsSchema)
+      const result = await validateRequest(
+        mockContext as unknown as Context,
+        ParamsSchema,
+      )
 
       expect(result.success).toBe(true)
       if (result.success) {
@@ -57,7 +61,10 @@ describe('validateRequest', () => {
         false,
       )
 
-      const result = await validateRequest(mockContext as any, ParamsSchema)
+      const result = await validateRequest(
+        mockContext as unknown as Context,
+        ParamsSchema,
+      )
 
       expect(result.success).toBe(false)
       if (!result.success) {
@@ -88,7 +95,10 @@ describe('validateRequest', () => {
         false,
       )
 
-      const result = await validateRequest(mockContext as any, QuerySchema)
+      const result = await validateRequest(
+        mockContext as unknown as Context,
+        QuerySchema,
+      )
 
       expect(result.success).toBe(true)
       if (result.success) {
@@ -100,7 +110,10 @@ describe('validateRequest', () => {
     it('should validate empty query parameters', async () => {
       const mockContext = createMockContext({}, {}, null, false)
 
-      const result = await validateRequest(mockContext as any, QuerySchema)
+      const result = await validateRequest(
+        mockContext as unknown as Context,
+        QuerySchema,
+      )
 
       expect(result.success).toBe(true)
       if (result.success) {
@@ -122,7 +135,10 @@ describe('validateRequest', () => {
       const requestBody = { name: 'John', email: 'john@example.com', age: 25 }
       const mockContext = createMockContext({}, {}, requestBody, true)
 
-      const result = await validateRequest(mockContext as any, BodySchema)
+      const result = await validateRequest(
+        mockContext as unknown as Context,
+        BodySchema,
+      )
 
       expect(result.success).toBe(true)
       if (result.success) {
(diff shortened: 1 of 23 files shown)


its-everdred:
Hi @daatsuka, I appreciate the contribution! Unfortunately this branch has fallen out of sync and we've decided to de-prioritize the associated issue. Thank you for your effort regardless!

PR #230, opened 2025-11-04
feat: add endpoint for getting op sepolia eth from faucet

tremarkley:
Closes https://github.com/ethereum-optimism/actions/issues/231

I deployed a OP Sepolia faucet that the actions demo app will hook into. When we are ready to integrate AAVE markets we will leverage the OP sepolia market and we will use this faucet to drip ETH to wallets on OP Sepolia, so that they can interact with the market.

PR #352, opened 2026-04-01
Expose public client for chain via Actions SDK

### LINKED ISSUE #232: Expose a way to get public client for a chain via the actions sdk
More context: https://github.com/ethereum-optimism/actions/pull/230#discussion_r2492258983

its-applekid:
## Summary

Closes #232

Exposes the internal `ChainManager.getPublicClient()` method through the `Actions` class, allowing SDK consumers to access viem `PublicClient` instances for making on-chain calls without creating their own clients.

## Motivation

As noted in #230 (discussion), developers currently need to manually create public clients with `createPublicClient()` to interact with on-chain contracts, even though the Actions SDK already manages properly configured clients internally. This leads to duplicated client creation logic and potentially misconfigured transports.

## Changes

- **`packages/sdk/src/actions.ts`**: Added `getPublicClient(chainId)` method that delegates to `ChainManager`
- **`packages/sdk/src/index.ts`**: Re-exported `PublicClient` type from viem for consumer convenience
- **`packages/sdk/src/__tests__/actions.test.ts`**: Added unit tests for the new method:
  - Returns `PublicClient` for configured chain
  - Throws error for unsupported chain
  - Returns same cached instance on repeated calls
  - Integration test: makes real on-chain call (external test)

## Usage

```typescript
const actions = new Actions(config, deps)

// Get a public client for Optimism
const publicClient = actions.getPublicClient(10)

// Use it for on-chain calls
const blockNumber = await publicClient.getBlockNumber()
const balance = await publicClient.getBalance({ address: '0x...' })
```

## Testing

- Build: ✅
- Lint: ✅ (0 errors)
- Tests: ✅ (442 passed, 11 skipped)


jefr90:
Thanks. Closing this for now: we haven't landed on a clear consumer use case for exposing the public client directly via the SDK. Happy to revisit when there's a concrete reason to surface it.

PR #356, opened 2026-04-02
Wallet Refactor: Support local EOAs and make embedded optional

its-everdred:
## Summary

Adds EOA support via `toActionsWallet` and makes the hosted wallet provider optional in `ActionsConfig`.

Agents and other consumers that only need a smart wallet can continue to use the existing `actions.wallet.createSmartWallet({ signer })` path — it already accepts any viem `LocalAccount` as the signer.

## Changes

- [x] `ActionsConfig.wallet.hostedWalletConfig` is now **optional**
- [x] `toActionsWallet` accepts a viem `LocalAccount` in addition to hosted provider params
- [x] `WalletNamespace` receives an `ActionsContext` at construction so the local branch uses its own deps (no reach into provider internals)
- [x] `LocalWallet` returned when `toActionsWallet` is called with a `LocalAccount`.

## Usage

```ts
import { createActions } from '@eth-optimism/actions-sdk/node'
import { privateKeyToAccount } from 'viem/accounts'

// Local-only (no hosted wallet provider configured)
const actions = createActions({
  chains: [...],
  wallet: {
    smartWalletConfig: { provider: { type: 'default' } },
  },
})

const account = privateKeyToAccount(process.env.AGENT_KEY as `0x${string}`)

// EOA via toActionsWallet
const eoaWallet = await actions.wallet.toActionsWallet(account)

// Smart wallet (existing path, unchanged)
const { wallet: smartWallet } = await actions.wallet.createSmartWallet({ signer: account })
```

Existing Privy/Turnkey consumers are unchanged. When a hosted provider is configured, `toActionsWallet` accepts both provider params and a `LocalAccount`; when it isn't, only `LocalAccount` is accepted (type-enforced).

## Follow-ups

All follow-on work is tracked under the umbrella wallet refactor issue #396, including the rename of the existing `EOAWallet` class (misnomer, the class doesn't require an EOA).


its-applekid (approved):

packages/sdk/src/actions.ts
```diff
@@ -229,7 +229,42 @@ export class Actions<
       SmartWalletProvider
     >
   > {
-    const hostedWalletProviderConfig = config.hostedWalletConfig.provider
+    const hostedWalletProvider = config.hostedWalletConfig
+      ? await this.createHostedWalletProvider(config.hostedWalletConfig)
+      : undefined
+
+    let smartWalletProvider: SmartWalletProvider
```
jefr90: I see this `let` wasn't introduced by this PR, tho I'm wondering if we could swap this to a `const` maybe through the use of an IFEE

packages/sdk/src/wallet/core/namespace/WalletNamespace.ts
```diff
@@ -10,6 +14,30 @@ import type { SmartWalletCreationResult } from '@/wallet/core/providers/smart/ab
...
+  }
+  if (
+    typeof record.signMessage !== 'function' ||
+    typeof record.signTransaction !== 'function'
+  ) {
+    return false
+  }
+  if ('signTypedData' in record && typeof record.signTypedData !== 'function') {
+    return false
+  }
```
jefr90: This is OK, just wondering if we could check all in a one-liner or under a single condition? Since they share same return `false`, I feel it'd make this more readable

jefr90 (approved):
Couple nits, rest LGTM

packages/sdk/src/actions.ts
its-everdred: agreed, updated

packages/sdk/src/wallet/core/namespace/WalletNamespace.ts
its-everdred: good call! updated

PR #357, opened 2026-04-02
feat(sdk): ENS resolution support for recipient addresses

### LINKED ISSUE #214: Add ENS support on all relevant address fields
## Problem 

Home page example shows the ability to send assets to an ENS, but this isn't implemented yet

## Solution

Implement this.  End developer-facing functions like send should allow the passer to pass an ENS address instead of only an address. 

Open questions:
- [ ] which address fields should allow ENS? Only `send()` or more functions too?
- [ ] what should the interface be for this? Should this be an alternative to the `address` field where we enforce either address or ens is required? 
- [ ] what error states do we need to account for?

jefr90:
## Summary

Adds ENS name resolution support to the SDK so callers can pass either a hex address or an ENS name (e.g. `"vitalik.eth"`) wherever a recipient address is accepted. Also introduces a dedicated `EnsNamespace` with profile lookup and a `NameServiceProvider` interface as groundwork for future name services (Basename, Lens, etc.).

## Changes

**`src/ens/types.ts`**
- `EnsName` type and `isEnsName` guard
- `EnsInfo` interface — all standard ENSIP-5/18 text record fields (`avatar`, `display`, `description`, `url`, `email`, `keywords`, `twitter`, `github`, `discord`, `reddit`), all nullable
- `NameServiceProvider` interface — `getAddress`, `getName`, `getInfo`; designed for future providers to implement

**`src/ens/EnsNamespace.ts`** — implements `NameServiceProvider`
- `getAddress(input)` — resolves an ENS name or hex address to a checksummed hex address
- `getName(address)` — reverse-resolves an address to its primary ENS name, or `null`
- `getInfo(input)` — fetches all standard ENSIP-5/18 text record fields in a single batched call via `Promise.all`; coalesced into one Multicall3 `eth_call` when multicall batching is enabled on the client
- All three methods are TTL-cached (default 5 min)
- Falls back to a public Cloudflare mainnet RPC automatically if mainnet is not in the caller's chain config — ENS works even in L2-only setups

**`src/swap/namespaces/BaseSwapNamespace.ts`** + **`WalletSwapNamespace.ts`**
- ENS resolution lifted out of `SwapProvider` and into the namespace layer
- `RecipientResolver` callback injected into namespace constructors; `Actions` and `Wallet` wire `ens.getAddress` as the resolver
- `SwapProvider.getQuote` / `execute` now accept resolved types only (`SwapQuoteParamsResolved`, `SwapExecuteParamsResolved`) — providers never see an ENS name

**`src/swap/core/SwapProvider.ts`**
- Removed `resolveRecipient` private method and all ENS imports
- `getQuote(params: SwapQuoteParamsResolved)` and `execute(params: SwapExecuteParamsResolved | SwapQuote)` — only accept pre-resolved types

**`src/types/swap/base.ts`**
- `WalletSwapParams.recipient` widened from `Address` to `Address | EnsName`
- `SwapExecuteParamsResolved` and `SwapQuoteParamsResolved` moved here from `SwapProvider.ts` for easier standalone imports

**`src/utils/ens.ts`**
- `RecipientResolver` type and `passthroughResolver` moved here from `BaseSwapNamespace` so lend/wallet namespaces can reuse them

**`src/utils/validation.ts`**
- `validateRecipient` extracted as a stateless utility (was a private method on `SwapProvider`)

**`src/index.ts`**
- `EnsNamespace`, `EnsInfo`, `NameServiceProvider`, and error classes exported from the public SDK API

## Design notes

**No new dependencies** — ENS resolution uses `normalize` from `viem/ens` and viem `PublicClient` actions. viem is already a core SDK dependency.

**No breaking changes** — `Address` is a subtype of `Address | EnsName`, so all existing callers passing hex addresses continue to work without changes.

**Public RPC fallback** — if mainnet (chain ID 1) is not in the caller's config, the namespace automatically falls back to `https://cloudflare-eth.com` so ENS works in L2-only setups without any configuration changes.

**Extensible by design** — `EnsNamespace` implements `NameServiceProvider`. Future providers (Basename, Lens, Unstoppable Domains) implement the same interface and can be registered under their own namespace.

**Providers only see addresses** — ENS resolution is a namespace-layer concern. `SwapProvider` implementations receive `Address` only, keeping them clean of name-service logic.

## Error handling

| Scenario | Error |
|---|---|
| Invalid ENS name (fails normalisation) | `EnsResolutionError` |
| ENS name resolves to null or zero address | `EnsResolutionError` |
| RPC failure | `EnsRpcError` (with `cause` attached) |

## Test plan

- [x] `src/ens/EnsNamespace.spec.ts` — 16 unit tests covering all methods, caching, and error cases
- [x] `src/utils/ens.test.ts` — 12 unit tests covering the low-level utility
- [x] ENS resolution tests removed from `SwapProvider.test.ts` (responsibility moved to namespace layer)
- [x] All 54 test files pass
- [x] TypeScript typecheck passes clean
- [x] Build passes for SDK, backend, and frontend

Closes #214


packages/sdk/src/ens/EnsNamespace.ts
```diff
@@ -0,0 +1,158 @@
...
+   *
+   * If you need to resolve outside of an {@link Actions} instance (e.g. in a
+   * provider or script), use the lower-level `resolveAddress` utility instead.
+   * @param input - Hex address (0x...) or ENS name (e.g. "vitalik.eth")
+   * @returns Resolved hex address
+   * @throws {EnsNotConfiguredError} If mainnet is not in your chain configuration
+   * @throws {EnsResolutionError} If the name cannot be resolved
+   * @throws {EnsRpcError} If the RPC call fails
+   */
+  async resolve(input: Address | EnsName): Promise<Address> {
```
its-everdred: 🤔 the end use dev will invoke these functions via `actions.ens._____`, how do you feel about `actions.ens.getName(address)` and `actions.ens.getAddress(name)` for clarity? it also matches some of our other functions like `actions.swap.getQuote()`

packages/sdk/src/ens/EnsNamespace.ts
```diff
@@ -0,0 +1,158 @@
...
+   * Look up a text record for an ENS name or address.
+   * Common keys: "avatar", "url", "email", "com.twitter", "com.github"
+   * @param input - Hex address (0x...) or ENS name
+   * @param key - Text record key
+   * @returns Text record value, or null if not set
+   * @throws {EnsNotConfiguredError} If mainnet is not in your chain configuration
+   * @throws {EnsResolutionError} If the name cannot be resolved or normalized
+   * @throws {EnsRpcError} If the RPC call fails
+   */
+  async lookupText(
```
its-everdred: ohhh I like this function, I hadn't considered the additional info, but it's pretty useful! two thoughts:

1. instead of the dev having to request each key individually, what if we just had it return all ENS record keys in one object. Is there a way to do that with Viem that doesn't spam the node?
2. in the same vein as the comment above, how do we feel about renaming this something like `getInfo()` or `getMetadata()` or something?

packages/sdk/src/ens/EnsNamespace.ts
```diff
@@ -0,0 +1,158 @@
...
+        })) ?? null
+    this.textCache.set(cacheKey, {
+      value,
+      expiresAt: Date.now() + this.cacheTtlMs,
+    })
+    return value
+  }
+
+  private requireMainnetClient() {
+    const client = this.chainManager.tryGetPublicClient(mainnet.id)
```
its-everdred: This limitation makes sense but has the downside of the dev needing to fully support mainnet in their ActionsConfig and therefore support all actions (lend, borrow, swap, pay) on mainnet. 

what if instead we default to their mainnet RPC, and if they haven't configured one, we add a public RPC as a fallback?  that way devs still get ENS support even when they only support an L2?

its-everdred (commented):
I created an [issue](https://github.com/ethereum-optimism/actions/issues/371) for this, so lets punt for now:  
A half-baked thought and I'm not sure how it might change the interface:

What would it look like for Actions to be name service agnostic? Ex: what if a dev wanted to support base name service as well as ENS? One option would be `wallet.basename.getName()` and/or something like  `wallet.address.getInfo()` that's generic and fetches all know info, name services, ens keys, farcaster username, safes they're on, etc.

definitely not something we would need to add support for now, but if we like the idea, we might consider future proofing the interface now.

packages/sdk/src/ens/EnsNamespace.ts
jefr90: Good call, added public node as fallback

packages/sdk/src/ens/EnsNamespace.ts
jefr90: Done and done ✅

packages/sdk/src/ens/EnsNamespace.ts
jefr90: Good call, changed wording to be more inline with repo standards

its-everdred (commented):
🤔 One of the original intentions for this work was to have the `wallet.sendTokens(...)` params accept ENS, but I just looked at that code and realized there's a larger refactor we need to make there.  Let's punt that for now, I just created an [issue](https://github.com/ethereum-optimism/actions/issues/370) for it.

its-everdred (commented):
comment removed, github isnt letting me delete this

packages/sdk/src/swap/core/SwapProvider.ts
```diff
@@ -407,6 +426,17 @@ export abstract class SwapProvider<
...
+    recipient: Address | EnsName | undefined,
+  ): Promise<Address | undefined> {
+    return recipient
+      ? resolveAddress(
+          recipient,
+          this.chainManager.tryGetPublicClient(mainnet.id),
+        )
+      : undefined
+  }
+
```
its-everdred: 🙌  Good call having resolution happen in the base provider classes (e.g. SwapProvider) so concrete implementations like Uniswap and Velodrome never see or care about name resolution, they just receive a resolved Address.

🤔 Instead of defining duplicate `resolveRecipient` and `validateRecipient` functions in each provider base class (lend, swap, borrow, wallet), would we be able to even DRY these as stateless utils and import them as needed?

We could also make the `chainManager` optional in these new utils so providers like WalletNamespace can use the fallback util without needing a chainManager passed through.

packages/sdk/src/utils/ens.test.ts
```diff
@@ -0,0 +1,103 @@
+import type { Address, PublicClient } from 'viem'
+import { describe, expect, it, vi } from 'vitest'
+
+import { resolveAddress } from '@/utils/ens.js'
+
+const REAL_ADDRESS = '0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045' as Address
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000' as Address
```
its-everdred: I know this doesnt matter as a test, but fwiw, viem has this as an export: `import { zeroAddress } from 'viem'`

its-everdred (approved):
just [this issue](https://github.com/ethereum-optimism/actions/pull/357/changes#r3082924968) then looks good!

packages/sdk/src/swap/core/SwapProvider.ts
```diff
@@ -57,6 +60,16 @@ const BPS_DENOMINATOR = 10000n
...
 
+/** SwapExecuteParams with recipient narrowed to Address after ENS resolution */
+type SwapExecuteParamsResolved = Omit<SwapExecuteParams, 'recipient'> & {
+  recipient?: Address
+}
+
+/** SwapQuoteParams with recipient narrowed to Address after ENS resolution */
+export type SwapQuoteParamsResolved = Omit<SwapQuoteParams, 'recipient'> & {
+  recipient?: Address
+}
```
its-everdred: one other annoying nit:
do you mind moving these types to `src/sdk/ens/types.ts`? this will make it easier for sdk consumers to import and use these types separately from the full sdk. not sure if anyone would do that, but itll save us a refactor if someone does.

packages/sdk/src/utils/ens.test.ts
jefr90: Nice catch, I prefer to lean on correctness even in tests

packages/sdk/src/swap/core/SwapProvider.ts
jefr90: Done, moved to ens types

PR #426, opened 2026-04-24
fix: default swap execute recipient to wallet address

its-everdred:
## Summary
`WalletSwapNamespace.execute()` forwards `params.recipient` to the pass-through ENS resolver without defaulting to the wallet address. When a caller omits `recipient`, the resolver returns `undefined`, reaches downstream viem address helpers, and throws `InvalidAddressError: Address "undefined" is invalid`.

Both `getQuote` and `getQuotes` already default `recipient` to `this.wallet.address` — `execute` was missing the same fallback.

## Fix
Apply `params.recipient ?? this.wallet.address` before `resolveRecipient` in the raw-params branch of `execute()`.

## Test plan
- [x] Existing `WalletSwapNamespace` tests pass
- [x] `execute()` with no `recipient` succeeds (uses `wallet.address`)
- [x] `execute()` with an explicit `recipient` still honors the provided value


jefr90 (approved):

jefr90:
Would it be useful to add a test for this?

PR #428, opened 2026-04-27
feat(sdk): configurable approval-amount strategy

its-everdred:
## What

Adds an opt-in `approvalMode: 'exact' | 'max'` knob to swap and lend.

- `'exact'` (default): approve only what's needed for the operation.
- `'max'`: approve `MaxUint256` once, skip per-call approvals on recurring use (~+1 block + ~50k gas saved per recurring action on L2s).

Resolution: per-call → per-provider config → shared settings → `'exact'`. Mirrors how `defaultSlippage` resolves.

**Lend approval construction also moved from individual providers into the base provider.** Aave and Morpho source files no longer build approvals or reference `ApprovalMode` — they return a single-shape `LendOpenPosition` describing what the deposit needs, and the base does the rest. New lend protocols just describe; they don't have to re-implement approval logic.

**Base provider constructors are now `protected`.** `LendProvider` and `SwapProvider` are abstract, so user code shouldn't have been instantiating them directly — `protected` makes that contract explicit. Concrete providers (Aave, Morpho, Uniswap, Velodrome) declare their own public constructors that pass through to `super(...)`. No behavior change for the `Actions` orchestrator that wires them up.

## Two design choices worth flagging

**Swap providers keep building approvals.** The three shapes (Uniswap Permit2 outer+inner, Velodrome v2 plain ERC-20, Velodrome universal-router with `transferFrom`) are too different to centralize without leaking Permit2 specifics into the base. Base provides `resolveApprovalMode` + `resolveErc20ApprovalAmount` + `resolvePermit2ApprovalAmount` helpers; providers compose them.

**Lend providers go declarative.** `_openPosition` returns a single-shape `LendOpenPosition`:

```ts
interface LendOpenPosition {
  assetAddress: Address
  spender?: Address  // required for ERC-20; omit for native
  transaction: TransactionData
  apy: number
}
```

The base reads `params.asset.type` (the source of truth for native vs ERC-20) and:
- **Native:** sends ETH inline as `msg.value`, no approval emitted.
- **ERC-20:** resolves `approvalMode`, builds `approve(spender, amount)`. Throws with a clear error if the provider forgot `spender`.

No discriminator on the position type itself — the base derives intent from `params.asset` instead of having every provider re-declare it. Aave's WETHGateway native-ETH path and standard ERC-20 deposits both flow through the same return shape.

## Tests

- 507 ✓ / 10 skipped / 0 failed
- New: `resolveApprovalMode` resolution chain (4 cases: per-call wins, falls through provider, falls through global, default)
- New: Velodrome regression — `'max'` mode emits `approve(maxUint256)`; `'exact'` (default) emits `approve(amountIn)`
- New: chokepoint guard test — `buildSwapTransactions` throws if `_getQuote` forgot to populate `recipient`
- Aave native-ETH "no approval" path still passes (base correctly skips approval based on `params.asset`)
- 0 lint errors

End-to-end "max-then-skip" verification will land alongside the CLI changes that consume this branch.


packages/sdk/src/actions/lend/core/LendProvider.ts
```diff
@@ -254,6 +287,38 @@ export abstract class LendProvider<
...
   }
 
+  /**
+   * Build an ERC20 approval transaction
+   * @param tokenAddress - Address of the token to approve
+   * @param spender - Address to approve spending for
+   * @param amount - Amount to approve
+   * @returns Transaction data for the approval
+   */
+  protected buildApprovalTx(
```
its-everdred, line 315: this name is too vague

packages/sdk/src/actions/lend/core/LendProvider.ts
```diff
@@ -97,11 +106,35 @@ export abstract class LendProvider<
...
+    return {
+      amount: amountWei,
+      asset: position.asset,
+      marketId: params.marketId.address,
+      apy: position.apy,
+      transactionData: {
+        ...(approval ? { approval } : {}),
+        position: position.transaction,
+      },
+    }
```
its-everdred: it's hard to read everything this function is doing, may be worth breaking out any util functions. if they're shareable between other actions like swap or borrow, lets pull it out of lend into the shared folder

packages/sdk/src/actions/lend/core/LendProvider.ts
```diff
@@ -254,6 +287,38 @@ export abstract class LendProvider<
...
+    approvalMode?: ApprovalMode
+  }): ApprovalMode {
+    return (
+      params.approvalMode ??
+      this._config.approvalMode ??
+      this._settings.approvalMode ??
+      'exact'
+    )
+  }
+
```
its-everdred: This ordering will be true of Lend, Swap, and Borrow providers. see if we can pull it into the BaseProvider class or some other abstraction

packages/sdk/src/actions/lend/core/LendProvider.ts
```diff
@@ -273,31 +338,30 @@ export abstract class LendProvider<
...
-  }
 
   /**
-   * Abstract methods that must be implemented by providers
+   * Chain IDs supported by the underlying protocol.
+   * @description Each provider implements this to declare the chains its protocol
+   * is deployed on, without any SDK-level or developer-config filtering.
+   * @returns Array of chain IDs the protocol natively supports
    */
+  abstract protocolSupportedChainIds(): number[]
```
its-everdred, line 370: is this more accurate if it were  providerSupportedChainIds?

packages/sdk/src/actions/swap/core/SwapProvider.ts
```diff
@@ -568,8 +626,14 @@ export abstract class SwapProvider<
...
   /**
    * Build provider-specific approval transactions for a swap.
+<<<<<<< kevin/approval-mode
+   * Called by the base class with a validated recipient and resolved approvalMode.
+   * Implementations read `quote.approvalMode` to choose between exact and max approvals.
+   * @param quote - SwapQuote with recipient and approvalMode set
+=======
    * Called by the base class during executeFromQuote with a validated recipient.
    * @param quote - NormalizedSwapQuote with recipient guaranteed to be set
+>>>>>>> main
```
its-everdred: remove this conflict

packages/sdk/src/actions/swap/core/SwapProvider.ts
```diff
@@ -383,10 +432,15 @@ export abstract class SwapProvider<
...
    * Used by both the quote-execute path and provider _execute implementations.
+<<<<<<< kevin/approval-mode
+   * Reads `quote.approvalMode` (populated by `execute()` at entry).
+   * @param quote - SwapQuote with recipient and approvalMode set
+=======
    * Normalises `quote.recipient` to `quote.quotedRecipient` when absent so
    * provider `_buildApprovals` implementations can dereference `quote.recipient`
    * without per-provider fallback logic.
    * @param quote - SwapQuote (recipient may be unset; defaults to quotedRecipient)
+>>>>>>> main
```
its-everdred: remove

its-everdred (commented):
nits

packages/sdk/src/actions/lend/core/LendProvider.ts
```diff
@@ -103,11 +114,34 @@ export abstract class LendProvider<
...
+    // ERC-20 deposits resolve approval mode and build an approve(spender, amount) tx.
+    let approval: TransactionData | undefined
+    if (!isNativeAsset(params.asset)) {
+      const approvalMode = resolveApprovalMode(
+        params.approvalMode,
+        this._config.approvalMode,
+        this._settings.approvalMode,
+      )
+      approval = this.buildLendApproval(position, approvalMode, amountWei)
+    }
```
jefr90: We could turn this let into a const here using a ternary, correct? Small nit, anti let bias

packages/sdk/src/types/actions.ts
```diff
@@ -114,6 +137,22 @@ export interface ActionsContext {
...
+ * Each subsequent operation needs its own approval transaction.
+ *
+ * `"max"` approves `maxUint256` for ERC-20 spenders / `maxUint160` for
+ * Permit2's inner allowance. Subsequent operations skip the re-approval
+ * round trip until the underlying allowance is consumed or expires.
+ *
+ * Default is `"exact"` for safety. Demo / dogfood configs typically opt into
+ * `"max"` to avoid an extra approval tx per swap or supply.
+ */
+export type ApprovalMode = 'exact' | 'max'
```
jefr90, line 154: Wondering if there's any upside to swapping this out for an enum and referencing like `ApprovalMode.Max` rather than referencing the raw string

jefr90 (approved):
1 nit 1 q, rest lgtm

packages/sdk/src/types/actions.ts
its-everdred: ohh good call

packages/sdk/src/actions/lend/core/LendProvider.ts
its-everdred: ah good catch ty

packages/sdk/src/types/actions.ts
its-everdred: oh interesting claude take:

> the enum vs string-union question on ApprovalMode. My take: keep the string union. every other discriminator/mode in this codebase (SwapProviderName, LendProviderName, routerType, etc.) is a string union, the TS team has moved away from recommending enums (runtime emission, tree-shaking quirks), and the codebase already gets autocomplete from the literal types. Switching just ApprovalMode would be inconsistent for marginal gain.

jefr90: That makes sense to me @its-everdred

jefr90 (approved):

PR #430, opened 2026-04-27
CLI + SDK Improvements

its-everdred:
## Summary

Three SDK performance wins that cut a USDC→OP swap on Base Sepolia from ~18s to ~6-9s wall time, plus a CLI demo config tweak so recurring swaps skip the per-call approval cost. A handful of incidental SDK cleanups along the way.

### Performance

- **Drop redundant double-wait in `EOAWallet.sendBatch`**. Was waiting for 2 confirmations between txs; one is enough. Saves a full block-time per sub-tx in a multi-tx batch.
- **Lower receipt polling for fast L2s**. viem's default `pollingInterval` is 4000ms (sized for L1 mainnet). Default to 1500ms for fast chains; mainnet stays at 4000ms. Configurable per-chain. Bundler client shares the same default.
- **Attach a nonce manager to the EOA wallet**. Avoids nonce contention when the same signer dispatches concurrent txs.
- **Demo CLI opts into `approvalMode: 'max'`** for both lend and swap settings in `packages/cli/src/demo/config.ts`. Subsequent swaps for the same token pair skip Permit2 + ERC-20 approval txs entirely.

### SDK / demo cleanup

- New SDK helpers: `getLendMarketAllowlist`, `Wallet.has(namespace)` capability check, `CHAIN_SHORTNAMES` constant.
- Barrel exports: `LEND_ACTIONS`, `LendProviderName`, `ApprovalMode` / `APPROVAL_MODES`.
- `getMarket` now takes the market directly.
- Demo markets switched to `getAssetAddress`.
- Lend help examples derived from the live config instead of hardcoded.
- Lend providers iterator narrowed to skip the `settings` key.
- Demo config guarded against mainnet; unichain dropped from the demo chain set and CLI resolver.
- `format DeployMorphoMarket prettier drift`.

## Test plan

- [x] SDK unit tests pass
- [x] CLI tests pass (195 / 195)
- [x] Backend + frontend tests pass
- [ ] Local CLI smoke: USDC→OP swap on Base Sepolia, first run includes approvals, second run is approval-free
- [ ] Verify wall-time reduction vs \`feat/cli-swap\` baseline

## Notes

- Targets \`feat/cli-swap\` (the upstream CLI base) rather than \`main\` so it can land alongside the CLI work.
- The Velodrome universal-router EOA fix and Uniswap v4 exact-out action-byte fix have already shipped via separate PRs (#441 et al.); duplicate commits will drop out when this branch is rebased after \`feat/cli-swap\` next merges \`main\`.


packages/sdk/src/wallet/core/wallets/eoa/EOAWallet.ts
```diff
@@ -83,12 +94,6 @@ export abstract class EOAWallet extends Wallet {
     const receipts: EOATransactionReceipt[] = []
     for (const tx of transactionData) {
       const receipt = await this.send(tx, chainId)
-      const publicClient = this.chainManager.getPublicClient(chainId)
-      // wait an extra confirmation so give time for nonce to be updated
-      await publicClient.waitForTransactionReceipt({
-        hash: receipt.transactionHash,
-        confirmations: 2,
-      })
```
its-everdred, line 91: safe to remove the `waitForTransactionReceipt` altogether because `this.send()` already awaits confirmation.

jefr90 (approved):
🔥 lgtm

PR #434, opened 2026-04-28
fix(sdk): require strict quote recipient match on execute

its-everdred:
There are two ways to generate a swap quote:

- \`actions.swap.getQuote(...)\` — pricing/routing only, no wallet bound. \`recipient\` defaults to the Universal Router \`msg.sender\` sentinel. Useful for general pricing outside the context of a specific wallet making a trade.
- \`wallet.swap.getQuote(...)\` — bound to a wallet. \`recipient\` is set to that wallet's address.

Either can be passed to \`wallet.swap.execute(quote)\`. Today, when the quote's recipient differs from the executing wallet, \`execute\` silently overrides the recipient and re-encodes calldata.

## Problem

The silent override is dangerous on Velodrome v2 / leaf routers, where the recipient is encoded **directly** into \`swapExactTokensForTokens(amounts, route, recipient, deadline)\`. If re-encoding doesn't fire (or doesn't rewrite the right field), the literal sentinel address ends up on chain as the recipient and tokens are lost. Beyond that specific bug, the contract — \"the calldata you read is not necessarily the calldata that runs\" — is a long-term footgun.

## Solution

If a quote was generated by \`actions.swap.getQuote\` (or any path where \`recipient\` doesn't match the executing wallet), \`wallet.swap.execute\` throws. Re-quote via \`wallet.swap.getQuote\` to bind the quote to the wallet.

\`SwapQuote.recipient\` becomes required (the dual \`recipient\` / \`quotedRecipient\` fields collapse into one). The \`NormalizedSwapQuote\` alias and the calldata re-encoding branch in \`SwapProvider.executeFromQuote\` are removed.

### DIFF
--- a/packages/sdk/src/actions/swap/__mocks__/MockSwapProvider.ts
+++ b/packages/sdk/src/actions/swap/__mocks__/MockSwapProvider.ts
@@ -10,7 +10,6 @@ import type { Asset } from '@/types/asset.js'
 import type {
   GetSwapMarketParams,
   GetSwapMarketsParams,
-  NormalizedSwapQuote,
   ResolvedSwapParams,
   SwapMarket,
   SwapProviderConfig,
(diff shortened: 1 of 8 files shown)


jefr90 (approved):
lgtm

tremarkley (approved):

PR #443, opened 2026-04-29
fix(sdk): Velodrome universal router approvals for EOAs (#438)

### LINKED ISSUE #438: Velodrome universal router approval shape breaks for EOA wallets
## Summary

`VelodromeSwapProvider`'s approval builder emits a direct `IERC20.transfer(router, amount)` instead of `approve()` when `routerType === 'universal'`, citing 4337 atomic batching. EOA wallets cannot batch atomically; the swap reverts onchain with `TRANSFER_FAILED`.

`routerType: 'universal'` is currently configured only for Base Sepolia (`packages/sdk/src/actions/swap/providers/velodrome/addresses.ts:75`), so this is the chain where the bug surfaces today.

## Reproduction

Branch: `kevin/swap-perf` @ `12d5f6ed`
Wallet: EOA (CLI `--private-key` flow, no smart-account)
Chain: Base Sepolia

```sh
actions wallet swap execute \
  --in USDC_DEMO --out OP_DEMO \
  --amount-in 1 \
  --chain base-sepolia \
  --provider velodrome
```

## Result

`exit=5`, onchain revert:

```json
{
  "error": "Execution reverted with reason: TRANSFER_FAILED.\n...\n  to:     0x4b94B729d6183c9efD0071f0790e984bAF46E093",
  "code": "onchain",
  "retryable": false,
  "details": {
    "cause": {
      "errorName": "TransactionExecutionError",
      "shortMessage": "Execution reverted with reason: TRANSFER_FAILED."
    }
  }
}
```

`0x4b94B729d6183c9efD0071f0790e984bAF46E093` is the configured Velodrome universal router on Base Sepolia.

The same pair via Uniswap (`--provider uniswap`) succeeds: 1 USDC → ~5.498 OP.

## Initial diagnosis (superseded — see Root cause below)

Original analysis attributed the bug to a missing wallet-kind signal: smart wallets batch atomically and can pre-`transfer`, EOAs can't. The implied fix was to thread an `atomicBatching: boolean` through swap params/quote and have Velodrome branch on it.

A research spike on the actual contract (see comment thread) showed this framing was wrong. The bug is upstream — we're calling the contract with the wrong selector — and the fix doesn't need any wallet-kind plumbing.

## Root cause

The Velodrome universal router on Base Sepolia (`0x4b94B729d6183c9efD0071f0790e984bAF46E093`) is an unverified deployment of [velodrome-finance/universal-router](https://github.com/velodrome-finance/universal-router), a fork of Uniswap's Universal Router. Same `execute(commands, inputs, deadline)` shape, same `payerIsUser` flag, same Permit2 fallback.

The contract has a first-class EOA path. From [`Permit2Payments.payOrPermit2Transfer`](https://github.com/velodrome-finance/universal-router/blob/main/contracts/modules/Permit2Payments.sol):

```solidity
function payOrPermit2Transfer(token, payer, recipient, amount) {
    if (payer == address(this)) { pay(...); }                 // payerIsUser=false: use router's own balance
    else {
        try ERC20.transferFrom(payer, recipient, amount);     // payerIsUser=true: standard approval works
        if (failed) permit2TransferFrom(...);                 // Permit2 fallback if no allowance
    }
}
```

The `V2_SWAP_EXACT_IN` / `V3_SWAP_EXACT_IN` inputs include a `payerIsUser` boolean. With `payerIsUser=true` and a standard `IERC20.approve(router, amount)`, the router pulls tokens via plain `transferFrom` from `msg.sender` (whether EOA or smart wallet). No pre-deposit, no Permit2 signature, no smart-wallet batching required.

The SDK currently hardcodes `payerIsUser: false` in [`v2.ts:221`](https://github.com/ethereum-optimism/actions/blob/main/packages/sdk/src/actions/swap/providers/velodrome/encoding/routers/v2.ts#L221) and [`cl.ts:147`](https://github.com/ethereum-optimism/actions/blob/main/packages/sdk/src/actions/swap/providers/velodrome/encoding/routers/cl.ts#L147), then compensates with a pre-`transfer(router, amount)` in [`approval.ts`](https://github.com/ethereum-optimism/actions/blob/main/packages/sdk/src/actions/swap/providers/velodrome/encoding/routers/approval.ts). That pre-transfer is what breaks for EOAs (sequential dispatch: tokens at router but no allowance set, swap reverts with `TRANSFER_FAILED`). The encoder's JSDoc at `v2.ts:195` admits the smart-wallet-only assumption: *"tokens are pre-transferred to the Router by the smart wallet."*

We're calling the contract incorrectly, not missing a wallet-kind signal.

## Fix

Three small edits, no new abstractions, no plumbing through the SDK:

| File | Change | ~LOC |
|---|---|---|
| `velodrome/encoding/routers/v2.ts` | Flip `payerIsUser: false` → `true` (line 221). Update JSDoc on `encodeUniversalV2Swap` to document the standard `approve + transferFrom` flow. | 3 |
| `velodrome/encoding/routers/cl.ts` | Flip `payerIsUser: false` → `true` (line 147). Same JSDoc update. | 3 |
| `velodrome/encoding/routers/approval.ts` | Delete the `routerType === 'universal'` transfer-shortcut branch entirely. All router types route through `buildApprovalTxIfNeeded`. Drop the `routerType` parameter. | ~10 (mostly subtraction) |
| `velodrome/VelodromeSwapProvider.ts` | Drop the `routerType` argument from the `buildTokenApproval` call site. | 1 |
| **Source subtotal** | | **~17 LOC** |

### Effect on smart wallets

Equivalent or better. Old smart-wallet path was `transfer + execute(payerIsUser=false)` (2 ops in the UserOp). New path is `approve + execute(payerIsUser=true)` (also 2 ops). Same atomicity, same gas profile. Approve persists across swaps; the transfer shortcut did not, so the new path is actually more efficient for repeat swaps on the same pair.

### Test plan

- **Encoding regression** (`encoding.v2.test.ts`, `encoding.cl.test.ts`): decode the inner `V2_SWAP_EXACT_IN` / `V3_SWAP_EXACT_IN` input from the universal-router `execute()` and assert `payerIsUser === true`. Existing tests don't assert on this flag, so they stay green; we're tightening.
- **Approval regression**: on a universal-router chain, insufficient allowance → `tokenApproval` is `approve(router, amount)`, NOT `transfer(router, amount)`. Fails without the fix; passes with it.
- Existing `routerType: 'v2'` / `routerType: 'leaf'` encoding and approval tests continue to pass unchanged.

### Acceptance criteria

- [ ] CLI `actions wallet swap execute --in USDC_DEMO --out OP_DEMO --amount-in 1 --chain base-sepolia --provider velodrome` succeeds against an EOA. Output matches Uniswap's result for the same pair (~5.5 OP).
- [ ] `pnpm typecheck && pnpm lint && pnpm test` green; zero new warnings.
- [ ] Regression tests in place that fail without the fix.
- [ ] Changeset added (patch bump for `@eth-optimism/actions-sdk`).

### Considered and discarded

- **`atomicBatching: boolean` threading through `SwapQuote` / `SwapExecuteParams` / `ResolvedSwapParams`, with `WalletSwapNamespace` injecting `instanceof SmartWallet`.** Was the original plan, before the research spike. Unnecessary once we use the contract correctly — there's no actual divergence in approval shape between EOAs and smart wallets. Saves ~50 LOC of plumbing and avoids touching wallet-layer types right before #396 reshapes them.
- **Capability object on `Wallet`.** Same reason.
- **Permit2 integration.** The contract supports it (commands `0x02 PERMIT2_TRANSFER_FROM`, `0x0a PERMIT2_PERMIT`), but it's not needed to fix this. The `try transferFrom` first branch in `payOrPermit2Transfer` makes plain ERC20 approve work; Permit2 is a future optimization for gasless approvals.

### Open question

Why was `payerIsUser: false` baked into the encoders? Worth a brief look at git blame during implementation to confirm there isn't a deliberate rationale that broke when other chains were added.

### See also

- #396 — wallet-layer refactor that this fix is forward-compatible with (and that we no longer need to coordinate with for this fix)

its-everdred:
Closes #438.

## What's broken

Velodrome is one of this SDK's swap providers. On chains where it routes through the "Universal Router" contract (today: Base Sepolia), every swap fails for EOA wallets with `TRANSFER_FAILED`.

## Why

To pay for a swap, the SDK has to give the router access to the user's input tokens. Two standard ways:

1. **`approve`**: user grants the router permission to pull tokens later. Router calls `transferFrom` during the swap.
2. **`transfer`**: user moves tokens to the router up front; router consumes them from its own balance during the swap.

The SDK was using option 2. That works for smart wallets, which bundle the `transfer` and the swap into a single atomic on-chain action — if the swap reverts, the transfer reverts too, no funds are stuck.

EOAs can't bundle. They send the `transfer` and the swap as **separate transactions**. The first lands fine, the router has the tokens. The second tries to `transferFrom` again (the router's internal accounting expects it) and reverts because the user never approved anything. Result: tx fails, tokens sit at the router until someone sweeps them out.

## The fix

Switch to option 1 (the standard `approve` pattern). The Universal Router has a first-class flag for it (`payerIsUser: true`); we just need to set it. Now the SDK builds an `approve(router, amount)` followed by the swap. Works for EOAs (sequential txs, allowance set first), works for smart wallets (same two ops bundled atomically), no behavior change for callers.

## Code changes

- `encoding/routers/v2.ts`, `encoding/routers/cl.ts`: set `payerIsUser: true` in the swap calldata.
- `encoding/routers/approval.ts` → deleted: was building the bogus `transfer`. Now we use the same standard approval helper as everywhere else in the SDK (one `allowance` check, skip the `approve` if the user already has enough).
- 4 small follow-up commits from review: regression tests, schema dedup, encoder cleanup.

## Demo testing context

The demo wires Privy as the wallet provider with a bundler and paymaster configured **at the Privy dashboard, not in this codebase**. From the SDK's view it looks like a plain EOA, but each tx is actually routed through Privy's bundler as a UserOp from a smart-wallet contract. `msg.sender` at the router is whatever address Privy hands back, and the preceding `approve` sets allowance against that same address — so the fix is correct in that flow too. None of that wiring lives here, so only an end-to-end run can exercise it.

## Pre-merge checklist

- [x] CLI raw EOA: `actions wallet swap execute --in USDC_DEMO --out OP_DEMO --amount-in 1 --chain base-sepolia --provider velodrome` (the original #438 repro)
- [x] Demo (Privy + bundler/paymaster): USDC → OP swap on Base Sepolia
- 489 unit tests passing in the SDK; new regressions in `encoding.v2.test.ts`, `encoding.cl.test.ts`, and `VelodromeSwapProvider.routing.test.ts` lock in the fix.

## Out of scope

The Universal Router path also hardcodes the swap output recipient to a `msg.sender` sentinel, so a user-supplied `recipient` is silently ignored on these chains. Pre-existing, independent of this fix. Tracked in #444.


packages/sdk/src/actions/swap/providers/velodrome/encoding/routers/v2.ts
```diff
@@ -192,36 +202,29 @@ export function encodeSwap(params: EncodeSwapParams): Hex {
...
-      false, // payerIsUser
-      false, // isUni — false for Velodrome/Aerodrome
-    ],
-  )
+  const input = encodeAbiParameters(V2_SWAP_EXACT_IN_INPUT_PARAMS, [
+    UNIVERSAL_ROUTER_MSG_SENDER, // recipient = msg.sender (Universal Router sentinel)
+    params.amountInRaw,
+    params.amountOutMin,
+    routes,
+    true, // payerIsUser — router pulls from msg.sender via transferFrom
```
its-everdred, line 225: The core change is this line

jefr90 (approved):
lgtm

PR #445, opened 2026-04-29
refactor: introduce ActionsError base and migrate bare throws to named subclasses

### LINKED ISSUE #382: refactor: introduce ActionsError base and migrate bare throw new Error() to named subclasses
## Summary

The SDK today throws bare \`new Error(messageString)\` from ~8 validation
and provider call sites. The only named error class is
\`TransactionConfirmedButRevertedError\`. This forces callers to
message-string-match to discriminate failures — fragile for UI rendering,
opaque for monitoring, and not a pattern new contributors can follow.

Introduce a thin abstract base \`ActionsError\` that extends viem's
\`BaseError\`, then migrate every bare \`throw new Error(...)\` to a named
concrete subclass. The abstract base gives callers a single
\`instanceof ActionsError\` check to discriminate SDK-origin errors from
pure viem errors (RPC, ABI, transport) or unknown runtime errors.

### Shape

```ts
// packages/sdk/src/core/error/errors.ts
import { BaseError } from 'viem'

/**
 * Abstract base for all errors thrown from inside the SDK.
 *
 * Extends viem's BaseError to inherit rich metadata (\`shortMessage\`,
 * \`metaMessages\`, \`cause\`). Callers can discriminate SDK-origin errors
 * with a single \`instanceof ActionsError\` check, separate from:
 *   - viem's own errors (ABI, RPC, transport) — \`instanceof BaseError\`
 *   - unknown runtime errors — \`instanceof Error\`
 */
export abstract class ActionsError extends BaseError {}

export class ChainNotSupportedError extends ActionsError {
  override name = 'ChainNotSupportedError' as const
  constructor(
    public chainId: number,
    public supportedChainIds: readonly number[],
  ) {
    super(\`Chain \${chainId} is not supported\`, {
      metaMessages: [\`Supported chains: \${supportedChainIds.join(', ')}\`],
    })
  }
}

export class MarketNotAllowedError extends ActionsError {
  override name = 'MarketNotAllowedError' as const
  constructor(public address: \`0x\${string}\`, public chainId: number) {
    super(\`Market \${address} on chain \${chainId} is not in the allowlist\`)
  }
}
// …plus ~10 more
```

### Why an abstract base (not "each class extends BaseError directly")

- A single \`instanceof ActionsError\` check at the catch site distinguishes
  "this came from the SDK and I can offer a typed recovery" from "this is a
  generic viem / RPC / network failure I should treat as opaque."
- The base stays intentionally **empty** for now — no shared methods, no
  shared fields. Its job is just to be a stable anchor for the hierarchy.
  Future cross-cutting behavior (e.g. structured logging, i18n lookup,
  correlation IDs) can land on the base without touching every subclass.
- Still aligns with viem's idiom: we extend viem's \`BaseError\`. Subclasses
  are named concrete classes per failure mode. No \`code\` discriminator —
  the class name IS the discriminator.

### Callers narrow

Frontend:
```
try {
  await actions.lend.openPosition({ ... })
} catch (e) {
  if (e instanceof ChainNotSupportedError) {
    // typed SDK failure — offer targeted recovery
    showSwitchNetworkToast(e.supportedChainIds)
  } else if (e instanceof MarketNotAllowedError) {
    showToast('Select an allowed market')
  } else if (e instanceof ActionsError) {
    // any other SDK-origin error
    showToast(e.shortMessage)
  } else if (e instanceof BaseError) {
    // viem / RPC error
    showToast('Network error — try again')
  } else {
    // truly unknown
    showToast('Unknown error')
  }
}
```

## Current throw sites to migrate

\`packages/sdk/src/utils/validation.ts\`:
- \`validateAmountProvided\` → \`AmountRequiredError\`
- \`validateAmountPositiveIfExists\` → \`InvalidAmountError\`
- \`validateNotBothAmounts\` → \`ConflictingAmountsError\`
- \`validateNotSameAsset\` → \`SameAssetError\`
- \`validateNotZeroAddress\` → \`ZeroAddressError\`
- \`validateSlippage\` → \`SlippageOutOfRangeError\`
- \`validateChainSupported\` → \`ChainNotSupportedError\`
- \`validateAssetOnChain\` → \`AssetNotSupportedOnChainError\`

\`packages/sdk/src/utils/assets.ts\`:
- \`getAssetAddress\` → \`AssetNotSupportedOnChainError\` (reuse)

\`packages/sdk/src/lend/core/LendProvider.ts\`:
- \`validateConfigSupported\` → \`MarketNotAllowedError\`
- \`closePosition\` \`Asset metadata is required\` → \`AssetMetadataRequiredError\`
- \`getPosition\` \`marketId is required\` → \`MarketIdRequiredError\`

\`packages/sdk/src/swap/core/SwapProvider.ts\` and \`swap/namespaces/\`:
- Similar set around swap-specific validation and provider resolution

\`packages/sdk/src/lend/namespaces/\` and \`swap/namespaces/\`:
- \`No provider configured for...\` → \`ProviderNotConfiguredError\`
- Add more classes as discovered during the migration

Also: migrate \`TransactionConfirmedButRevertedError\` to extend \`ActionsError\`
so it joins the hierarchy.

## Scope

- 1 new abstract base \`ActionsError extends BaseError\` in \`core/error/\`
- ~10–12 new named concrete subclasses
- ~20 \`throw new Error(...)\` replacement sites
- Migrate existing \`TransactionConfirmedButRevertedError\` to extend \`ActionsError\`
- Update affected unit tests to match on \`instanceof\` + structured fields
  instead of message-string regex (e.g., \`.rejects.toThrow(ChainNotSupportedError)\`)
- Update demo frontend error-rendering to narrow via \`instanceof\`
- Update demo backend error logging to record \`error.name\` as a structured field

## Explicitly out of scope

- **No \`code\` discriminator field.** The class name IS the discriminator.
- **No rework of viem's error classes.** Our hierarchy extends viem's
  \`BaseError\`; viem-origin errors stay as-is and are caught via
  \`instanceof BaseError\`.

## Benefits

- Compile-time-safe error rendering in frontend (\`instanceof\` narrowing).
- Single \`instanceof ActionsError\` check separates SDK errors from viem
  errors at boundary code (logging middleware, global toast handler, etc.).
- Structured fields on every error enable correct recovery UX (user sees
  the actual allowed values, not regex-extracted from a string).
- Monitoring can index by \`error.name\` instead of regex on message.
- Clear pattern for every future thrown error in the SDK.
- Aligned with viem's idioms — our subclasses sit under viem's \`BaseError\`,
  not parallel to it.

## Costs

- Medium-size PR (~30–40 files).
- Any external consumer matching on error message strings breaks — but
  the ecosystem is small (demo backend/frontend only today).
- Coordination cost with in-flight branches that assume string-message
  errors; ideally landed during a quiet window.

## Acceptance criteria

- [ ] \`ActionsError\` abstract class lives in
      \`packages/sdk/src/core/error/errors.ts\` extending viem's \`BaseError\`.
- [ ] Zero bare \`throw new Error(...)\` in \`packages/sdk/src\` (except
      internal-bug / assertion cases that should never fire).
- [ ] Every current throw site has a dedicated named subclass of
      \`ActionsError\`.
- [ ] \`TransactionConfirmedButRevertedError\` extends \`ActionsError\`.
- [ ] Every error class has a unit test asserting:
  - Correct \`name\`
  - Structured fields are set
  - \`instanceof ActionsError\` returns true
  - \`instanceof BaseError\` returns true
- [ ] Demo backend logs \`error.name\` as a structured field.
- [ ] Demo frontend error rendering narrows via \`instanceof\` in at least
      the happy paths (validation, chain mismatch, market not allowed).
- [ ] Existing tests updated from message-regex matching to
      \`instanceof\`-based assertions.

## Related

- #380 — codify engineering principles (this refactor demonstrates several
  of them: "prefer viem patterns" by extending \`BaseError\`; "reuse before
  invention" by replacing scattered bare throws; abstraction hierarchy by
  sharing one common ancestor)

jefr90:
Resolves #382.

## Summary

- Adds `ActionsError extends BaseError` (from viem) as an abstract base class, plus 15 named concrete subclasses: `ChainNotSupportedError`, `MarketNotAllowedError`, `ProviderNotConfiguredError`, `MarketIdRequiredError`, `AmountRequiredError`, `InvalidAmountError`, `ConflictingAmountsError`, `SameAssetError`, `QuoteExpiredError`, `ExactOutputNotSupportedError`, `ZeroAddressError`, `SlippageOutOfRangeError`, `AssetNotSupportedOnChainError`, `NativeAssetAddressError`, `AssetMetadataRequiredError`
- Migrates every `throw new Error(...)` in `packages/sdk/src` to the appropriate named subclass (except intentional catch-all wrappers and exhaustiveness checks — see below)
- Promotes `TransactionConfirmedButRevertedError` to `extends ActionsError`
- Exports the full hierarchy from the SDK's public index
- Updates the demo backend to log `error.name` as a structured field; demo frontend to narrow errors via `instanceof` and display typed messages

## Intentional bare `Error` sites (not migrated)

| File | Site | Reason |
|---|---|---|
| `AaveLendProvider.ts`, `MorphoLendProvider.ts` | "Failed to open/close position" | Catch-all wrappers around contract execution failures; not user-fixable |
| `aave/sdk.ts`, `morpho/sdk.ts` | "Failed to get reserve/vault" | Same — catch-all wrappers |
| `velodrome/routers/v2.ts` | "Unknown router type" (×2) | Exhaustiveness checks (`never`) |
| `wallet/core/**` | Various internal wallet state assertions | Complex pre-existing wallet error model; out of scope |
| `utils/validateAddresses.ts` | "Invalid addresses found" | Internal startup validation, not a user-facing SDK error |

## Test plan

- [ ] New unit tests in `src/core/error/__tests__/errors.test.ts` cover every subclass (name, fields, `instanceof ActionsError`, `instanceof BaseError`)
- [ ] All existing tests updated to match new error messages — 58/58 test files pass
- [ ] `pnpm typecheck` — 0 errors
- [ ] `pnpm lint:fix` — 0 errors


packages/sdk/src/actions/lend/core/LendProvider.ts
```diff
@@ -86,7 +92,7 @@ export abstract class LendProvider<
    */
   async openPosition(params: LendOpenPositionParams): Promise<LendTransaction> {
     if (!params.walletAddress) {
-      throw new Error('walletAddress is required')
+      throw new ZeroAddressError('walletAddress')
```
its-everdred: 🤔 This isn't technically an issue with using the zero address, might be worth adding an `AddressRequiredError`.

jefr90: Nice catch, updated

packages/sdk/src/actions/lend/providers/aave/sdk.ts
```diff
@@ -95,14 +99,16 @@ export async function getReserve(
...
+      address: params.marketId.address,
+      chainId: params.marketId.chainId,
+      reason: 'Market not found in allowlist',
+    })
   }
 
   const addresses = getAaveAddresses(params.marketId.chainId)
   if (!addresses) {
-    throw new Error(`Aave V3 not deployed on chain ${params.marketId.chainId}`)
+    throw new ChainNotSupportedError(params.marketId.chainId, [])
```
its-everdred: 🤔 Do you mind if we use a single params obj with named attrs for all these errors, like we did above with `MarketNotAllowedError`?

packages/sdk/src/actions/lend/providers/morpho/sdk.ts
```diff
@@ -447,7 +452,10 @@ export async function findBestVaultForAsset(
   marketAllowlist: LendMarketConfig[],
 ): Promise<Address> {
   if (!marketAllowlist || marketAllowlist.length === 0) {
-    throw new Error('Market allowlist is required and cannot be empty')
+    throw new ProviderNotConfiguredError(
+      'marketAllowlist',
+      'Market allowlist is required and cannot be empty',
+    )
```
its-everdred: [Same](https://github.com/ethereum-optimism/actions/pull/445/changes#r3168930530) here, let's use named params so it's obvious what each string is used for.

packages/sdk/src/actions/swap/core/markets.ts
```diff
@@ -69,7 +70,11 @@ export function findMarket<T extends SwapMarketConfig>(
     )
     if (match) return match
   }
-  throw new Error(`Market with poolId ${poolId} not found on chain ${chainId}`)
+  throw new MarketNotAllowedError({
```
its-everdred: 🤔 I wonder if it's worth differentiating `MarketNotAllowedError` -- it's not in the allow list, or it's in the block list
from `MarketNotFoundError` -- we looked but can't find it

packages/sdk/src/actions/swap/core/SwapProvider.ts
```diff
@@ -468,9 +479,7 @@ export abstract class SwapProvider<
   private validateQuoteExpiration(quote: SwapQuote): void {
     const now = Math.floor(Date.now() / 1000)
     if (now >= quote.expiresAt) {
-      throw new Error(
-        `Quote expired at ${quote.expiresAt}, current time is ${now}`,
-      )
+      throw new QuoteExpiredError(quote.expiresAt, now)
```
its-everdred: [Same](https://github.com/ethereum-optimism/actions/pull/445/changes#r3168930530) here, let's use named params so it's obvious what each string is used for.

packages/sdk/src/core/error/errors.ts
```diff
@@ -0,0 +1,237 @@
...
+    this.address = params.address
+  }
+}
+
+export class ProviderNotConfiguredError extends ActionsError {
+  override name = 'ProviderNotConfiguredError' as const
+  provider: string
+
+  constructor(provider: string, details?: string) {
+    super(`Provider '${provider}' is not configured`, {
```
its-everdred:
```suggestion
    super(`A '${provider}' provider is not configured`, {
```
this: `A 'swap' provider is not configured` 
vs
orig: `Provider 'swap' is not configured`

Nit: This wording makes it a little more clear there can be many providers of each type, not just one

packages/sdk/src/core/error/errors.ts
```diff
@@ -0,0 +1,237 @@
...
+  override name = 'ExactOutputNotSupportedError' as const
+  provider: string
+
+  constructor(provider: string) {
+    super(`${provider} does not support exact-output swaps`, {
+      metaMessages: ['Provide amountIn instead of amountOut'],
+    })
+    this.provider = provider
+  }
+}
```
its-everdred, line 175: ℹ️ context: there's this quirk with velodrome where you can't ask the contract to purchase exactly X of the output token, although you can with Uniswap. looks good though

its-everdred (approved):
couple nits but looks good!

PR #383, opened 2026-04-16
Refactor: shared namespace foundations

its-everdred:
Foundation PR for borrow support (#388). Pure refactor + one bug fix — no new user-facing features.

## What landed

- [x] Shared namespace abstraction so Lend, Swap, and the upcoming Borrow namespace stop reimplementing the same provider registry and wallet dispatch logic
- [x] Morpho-specific contracts/types promoted to a shared home so a future Morpho borrow provider can reuse them
- [x] Common types (transaction options, asset/chain filter) promoted out of \`lend/\` since they aren't lend-specific
- [x] Type-safety tightening across the namespace layer (narrower chain-id typing, generalized provider routing, no unsafe casts, \`readonly\` on non-mutating params)
- [x] Shared allowlist lookup helper, deduping a pattern repeated across two call sites
- [x] **Bug fix**: lending position balance formatting hardcoded USDC decimals — broken for any non-USDC vault. Now derives decimals from the market's allowlist entry with an on-chain fallback
- [x] Directory restructure:
  - Introduce \`actions/\` (wraps lend + swap, anticipates borrow) and \`actions/shared/\` (cross-domain protocol code + \`BaseNamespace\`)
  - Relocate ens under \`services/nameservices/ens/\` (anticipating sibling name services)
  - Absorb wallet-only utilities (\`executeTransactionBatch\`, \`retryOnStaleRead\`, transaction-result errors) into \`wallet/core/\`
  - Delete \`src/core/\` entirely
  - Relocate lend-specific market utilities (including \`findMarketInAllowlist\`) into \`actions/lend/utils/\`
- [x] Test hygiene: relocate stray \`*.test.ts\` / \`*.spec.ts\` files into sibling \`__tests__/\` directories (services, ens, utils)
- [x] ABI cleanup: consolidate the scattered inline \`PERMIT2_ABI\` copies into \`utils/abi/permit2.ts\`; replace custom ERC-4626 fragments with viem's built-in \`erc4626Abi\`

## Testing

- [x] Typecheck + lint clean across all packages
- [x] Lint baseline ratcheted down 6 warnings; zero new warnings
- [x] 477 tests passing (up from 441); existing Lend + Swap tests pass unchanged
- [x] New tests for the shared abstractions and both paths of the decimal fix

## Breaking changes

- \`LendOptions\` type renamed to \`TransactionOptions\` and moved to the common types directory. Demo backend/frontend updated in the same commit.
[...]

PR #372, opened 2026-04-15
Parallelize tests

its-everdred:
## Summary

Enables parallel test execution and splits CI into per-package jobs for faster builds.

### Part 1: Vitest Parallelism ✅
- Removed `fileParallelism: false` from `vitest.config.ts`
- All 439 tests pass with parallelism enabled
- Test duration: 22.93s (down from ~90s sequential)

### Part 2: CI Job Splitting ✅
- Split monolithic `check` job into 5 parallel jobs:
  - `install-and-build` — shared setup step
  - `check-sdk` — typecheck, lint, test
  - `check-backend` — typecheck, lint, test
  - `check-frontend` — typecheck, lint, test
  - `check-contracts` — build, test
- Jobs run in parallel after shared install/build
- Backend/frontend depend on SDK build (workspace dep)

## Benefits

- ✅ **Faster local tests** — parallel file execution
- ✅ **Faster CI** — parallel package jobs
- ✅ **Better isolation** — failures easier to identify per package
- ✅ **Scalable** — adding packages won't slow down everything

## Testing

- ✅ All SDK tests pass with parallelism (439 tests, 22.93s)
- ✅ Verified no race conditions or flaky tests
- ✅ CI will verify parallel job execution

## Closes

Closes #333

PR #450, opened 2026-05-04
Bump sdk to 0.6.0

its-everdred:
PRs since last version bump:

Features / API additions
#451 — Add chainIds param to wallet.getBalance()
#428 — Configurable approval-amount strategy: callers can now choose between exact-amount and unlimited approvals when opening positions.
#356 — Wallet refactor: native support for local EOA wallets; the embedded (4337) wallet is now optional rather than required.
#383 — Shared namespace foundations: reorganizes shared internals to support multiple action domains (lend, borrow, swap) under a common surface.

Public-surface refactor

#445 — Introduces an ActionsError base class and migrates bare throws across the SDK into named subclasses, giving consumers structured errors to catch.

Fixes

#443 — Bug fix in VelodromeProvider impacting EOA swaps, which previously reverted with TRANSFER_FAILED because the encoder always pre-transferred tokens to the router.
#434 — Bug fix in swap execution where a quote’s recipient did not match the execute call’s recipient; the SDK now rejects the mismatch instead of silently routing funds to the wrong address.
#441 — Bug fix in UniswapV4Provider affecting exact-output single-hop swaps, which used the wrong action byte and produced invalid calldata.
#426 — Bug fix in swap execution for calls that omitted a recipient; the SDK now defaults to the wallet address instead of leaving it unset.

Tooling / infra (incidentally touched the SDK package)

#385 — Bump runtime to Node 22.14.0.
#372 — Parallelize the test suite.


jefr90 (approved):
LGTM

PR #466, opened 2026-05-12
Borrow PR #5: Demo frontend

its-everdred:
## Problem

The SDK now has Borrow support, but the demo has no borrow functionality.

## Solution

borrow-repay

- **Borrow tab**
  - lend-position-driven market selection + borrow / repay form
  - sticky health bar with directional barbershop-stripe delta
  - review modal with projection and danger / liquidation warnings
  - active positions table (Collateral | APY | Borrowed split) with APY tooltip

- **Lend tab updates**
  - withdraw routes through the review modal when the deposit is pledged as collateral
  - max clamp respects the safe-ceiling LTV of the pledged position

- **SDK**
  - fetch `rateAtTarget` from the IRM alongside the market read so `Market.borrowApy` returns a real APY
  - convert vault shares to underlying via `convertToAssets(1e18)` for accurate `collateralAmount`
  - gate that call on `collateralAsset.address !== collateralToken` so non-vault Morpho markets don't revert
  - optimistic position update from `receipt.positionAfter` + delayed RPC reconciliation to bridge Base Sepolia propagation lag

- **Contracts**
  - `DeployMorphoBorrowMarket.s.sol` now seeds ~80% utilization at deploy time so AdaptiveCurveIrm yields a non-zero rate
  - new one-off `SeedMorphoBorrowMarket.s.sol` does the same against an already-deployed market, idempotent via `position(id, msg.sender).borrowShares > 0`

- **Shared frontend pieces**
  - extracted `PositionsTable` chrome (used by Lent Balance + Active Positions)
  - `ModeToggle` generalized to arbitrary two-option pairs (lend/withdraw and borrow/repay)
  - `AmountInput` token chip with optional click-to-open selector
  - `reviewModalParts` (`AmountRow`, `DetailRow`, `FormattedAmount`) shared by swap, withdraw, and borrow / repay review modals

## Follow-up

- [ethereum-optimism/actions#482](https://github.com/ethereum-optimism/actions/issues/482) — Simplified PriceProvider: SDK contract + Chainlink-backed demo implementation + frontend `stubPriceUsd` retirement
- Withdraw-with-pledged-collateral is currently two sequential transactions (`withdrawCollateral` then `closePosition`); batching via `wallet.sendBatch` is a follow-up

## Validation

- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] `pnpm lint:fix`
- [x] `pnpm test`


packages/demo/frontend/src/api/borrowApi.ts
```diff
@@ -0,0 +1,404 @@
...
+  OP_DEMO: 0.1,
+  ETH: 3000,
+  WETH: 3000,
+}
+
+export function stubPriceUsd(symbol: string): number {
+  return (
+    STUB_PRICES_USD[symbol] ?? STUB_PRICES_USD[symbol.replace('_DEMO', '')] ?? 0
+  )
+}
```
its-everdred: Replaced by SDK `PriceProvider` + per-shape `collateralPrice` / `borrowPrice` on borrow read shapes — tracked in #482. Stays as-is for this PR.

packages/sdk/src/actions/borrow/namespaces/WalletBorrowNamespace.ts
```diff
@@ -37,6 +39,24 @@ export class WalletBorrowNamespace extends BaseBorrowNamespace {
...
+   * @throws If the underlying provider's RPC fetch fails.
+   */
+  async getPosition(
+    params: Omit<GetBorrowPositionParams, 'walletAddress'>,
+  ): Promise<BorrowMarketPosition> {
+    return this.getProviderForMarket(params.marketId).getPosition({
+      ...params,
+      walletAddress: this.wallet.address,
+    })
+  }
```
its-everdred, line 58: missed this in the SDK PR.

packages/demo/frontend/src/api/actionsApi.ts
```diff
@@ -144,12 +105,13 @@ class ActionsApiClient {
     }
   }
 
-  async openLendPosition(
+  private async lendMutation(
+    endpoint: string,
```
tremarkley: I wonder if instead of a string this should be an enum and then inside `lendMutation`, it can map to the correct endpoint? or is there a reason this should accept any string value?

tremarkley (approved):

jefr90 (approved):
🔥 good stuff

packages/demo/frontend/src/api/actionsApi.ts
its-everdred: good call, fixed

PR #467, opened 2026-05-12
fix(frontend): bind swap quotes to wallet

its-everdred:
# Problem
Frontend-hosted wallet swaps were failing immediately at execution time with a recipient mismatch error.

This started after [#434](https://github.com/ethereum-optimism/actions/pull/434), which changed `wallet.swap.execute(...)` to fail when a quote is bound to a different recipient instead of silently rewriting the recipient during execution.

The frontend wallet path was still generating swap quotes through the read-only actions namespace instead of the wallet namespace. That produced swap calldata bound to a placeholder recipient rather than the active wallet address, so execution was rejected before the transaction could be submitted.

The server-wallet path also dropped the selected swap provider during execution, which could cause quote selection and execution to diverge.

# Solution
Generate frontend-wallet swap quotes through `wallet.swap.getQuote(...)` so the recipient is bound to the connected wallet before execution.

Forward the selected provider through the server-wallet execute path so quote selection and execution stay aligned across wallet modes.

Added a regression test covering the frontend-wallet quote path.

### DIFF
--- a/packages/demo/frontend/src/components/earn/EarnWithFrontendWallet.tsx
+++ b/packages/demo/frontend/src/components/earn/EarnWithFrontendWallet.tsx
@@ -1,11 +1,7 @@
-import { encodeFunctionData, type Address } from 'viem'
 import type {
   Wallet,
-  SupportedChainId,
-  Asset,
   ReactProviderTypes,
 } from '@eth-optimism/actions-sdk/react'
-import { mintableErc20Abi } from '@/abis/mintableErc20Abi'
 import { getBlockExplorerUrl } from '@/utils/blockExplorer'
 import Earn from './Earn'
 import {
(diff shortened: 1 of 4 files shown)


packages/demo/frontend/src/components/earn/EarnWithFrontendWallet.tsx
```diff
@@ -47,96 +153,7 @@ export function EarnWithFrontendWallet({
...
-          const assets = actions.getSupportedAssets()
-          const assetIn = assets.find(
-            (a) => a.address[params.chainId] === params.tokenInAddress,
-          )
-          const assetOut = assets.find(
-            (a) => a.address[params.chainId] === params.tokenOutAddress,
-          )
-          if (!assetIn || !assetOut) return null
-
-          return await actions.swap.getQuote({
```
its-everdred, line 127: This line was the issue, instead we now do `wallet.swap.getQuote()` which generates the quote specific to the wallet.

tremarkley (approved):

Issue #468, opened 2026-05-18
Borrow: AaveBorrowProvider end-to-end


its-everdred:
## Goal

Add full Aave borrow support as a first-class `BorrowProvider`, parallel to the existing Morpho borrow implementation, and make it work end to end through the SDK, demo backend, and demo frontend.

Aave should be configurable as a borrow provider anywhere the borrow domain is already used today: SDK actions namespace, wallet namespace, backend endpoints, and the user-facing borrow tab.

> [!WARNING]
> Aave is much more restrictive than Morpho for demo-market construction. We cannot freely create arbitrary Aave markets for `USDC_DEMO`, `OP_DEMO`, or vault-share collateral on the demo testnet. The demo may need a clever emulation strategy: use real Aave testnet reserves where available, wrap/mock only the display and price-feed layer, seed supported assets through faucets, or add a local/demo adapter that exercises the SDK/backend/frontend Aave flow without pretending custom demo assets are real Aave reserves. Make this explicit in the implementation plan before wiring the frontend.

### 1. SDK provider

Add `AaveBorrowProvider` under the borrow domain.

Expected shape:

- Extend the existing `BorrowProvider` base class.
- Register `aave` in `BORROW_PROVIDER_NAMES` / `BorrowProviders`.
- Export `AaveBorrowProvider` from the public SDK entry points.
- Wire `borrowModule.buildProviders()` so `config.borrow.aave` constructs the provider.
- Reuse or hoist shared Aave protocol pieces from the lend provider where appropriate: addresses, Pool ABI, WETH gateway/native ETH behavior, reserve metadata, health/account reads.
- Keep provider-specific code thin: protocol calls, Aave-specific encoding, Aave-specific position math, and Aave-specific quirks only.
- Support the standard borrow surface: `getMarkets`, `getMarket`, `getPosition`, `getQuote`, `openPosition`, `closePosition`, `depositCollateral`, `withdrawCollateral`, and `repay`.
- Preserve base-class validation, amount handling, approvals, receipt shape, and namespace routing semantics.

### 2. Aave market and position semantics

Map Aave reserve/account data into the generic borrow types without leaking Aave internals into the public borrow surface.

Implementation needs to decide and document:

- How an Aave borrow market is identified in `BorrowMarketId`.
- How collateral asset + debt asset pairs are represented when Aave exposes reserves independently rather than Morpho-style isolated markets.
[...]

PR #489, opened 2026-06-01
AaveBorrowProvider SDK support

its-everdred:
## Problem

Morpho supports lend + borrow, but Aave supports lend only — there is no Aave borrowing.

## Solution

AaveBorrowProvider — SDK (#468)

- Aave V3 borrow provider, parallel to Morpho — pure real-Aave, no demo logic.
- Shared `actions/shared/aave/` (addresses + Pool ABI) reused by lend and borrow.
- `aave-v3` market variant: discriminated `BorrowMarketId`/`BorrowMarketConfig` + synthetic `computeAaveBorrowMarketId`.
- Reads via multicall (reserve/account data, oracle); writes `borrow`/`repay`/`supply`/`withdraw`, native ETH via WETH gateway, full-repay via `maxUint256`.
- `depositCollateral` supports `max`.
- Provider registered + public exports + changeset.

SDK suite green; repo-wide typecheck / lint / build clean.


packages/sdk/src/actions/borrow/core/internalParams.ts
```diff
@@ -82,7 +82,7 @@ export function buildDepositCollateralInternalParams(
     walletAddress: base.walletAddress,
     options: params.options,
     approvalMode: base.approvalMode,
-    amountWei: toAmountWei(
+    amount: toAmountWeiOrMax(
```
its-everdred: this PR also adds the ability to deposit all collateral via max instead of forcing the caller to check wallet balance first, and then pass the value. `max` already exists on `withdrawCollateral`

packages/sdk/src/actions/borrow/core/BorrowProvider.ts
```diff
@@ -67,6 +64,13 @@ export abstract class BorrowProvider<
     super(config, chainManager, settings)
   }
 
+  /**
+   * The `BorrowMarketId` discriminator this provider services. Lets the
+   * namespace route a market to its provider by kind without naming concrete
+   * providers, and is the fallback when a provider carries no market allowlist.
+   */
+  public abstract get marketKind(): BorrowMarketId['kind']
```
its-everdred, line 72: Not sure I love this approach, created an [issue](https://github.com/ethereum-optimism/actions/issues/493) to reconsider.

packages/cli/src/demo/config.ts
```diff
@@ -9,7 +9,7 @@ import { getDemoChains } from '@/demo/chains.js'
 import {
   AaveETH,
   GauntletUSDCDemo,
-  MorphoUSDCBorrowDemo,
+  MorphoUSDCBorrowOPDemo,
```
jefr90, line 12: Why the rename here? Just curious

packages/sdk/src/actions/borrow/core/BorrowProvider.ts
jefr90: What to you feel may be a better option here?

packages/sdk/src/actions/borrow/providers/aave/presentation.ts
```diff
@@ -0,0 +1,260 @@
...
+
+// Aave fixed-point units: ray 1e27, wad 1e18, bps 1e4 (= 100%).
+// https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/math/WadRayMath.sol
+const RAY = 10n ** 27n
+const BPS = 10000n
+
+/** Ray rate to decimal fraction (e.g. 0.035). */
+export function rayToFraction(ray: bigint): number {
+  // 1e6 fixed point preserves precision before Number().
+  return Number((ray * 1_000_000n) / RAY) / 1_000_000
```
jefr90: Should we also have constants for `1_000_000n` and `1_000_000`? Small nit

packages/sdk/src/actions/borrow/core/quote.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export interface AssembleBorrowQuoteArgs {
+  provider: BorrowProviderName
+  action: BorrowAction
+  recipient: Address
+  /** Converted position before the action, or `null` for a fresh open. */
+  positionBefore: BorrowMarketPosition | null
+  /** Converted position after the action lands on-chain. */
+  positionAfter: BorrowMarketPosition
+  quoteAmounts: { borrowAmountRaw?: bigint; collateralAmountRaw?: bigint }
```
jefr90: Since both attributes for `quoteAmounts` are optional, should `quoteAmounts` itself be optional as well?

jefr90 (approved):
Couple questions, rest LGTM

packages/cli/src/demo/config.ts
its-everdred: good q! Aave borrow markets can take in 1 input lend asset and use it as collateral to borrow N output assets.  Meanwhile, morpho-blue borrow markets are 1:1. This particular borrow market will only take in USDC and let you borrow OP, so i want to make that distinction clear.

packages/sdk/src/actions/borrow/core/quote.ts
its-everdred: Good catch! Kept the container required since every action emits at least one amount, but tightened the type to a union so at least one is now required at compile time, rather than allowing an empty {}.

packages/sdk/src/actions/borrow/providers/aave/presentation.ts
its-everdred: yes, good call, done

its-everdred:
> What do you feel may be a better option here?


@jefr90 so far we've been differentiating and determining market providers based on the shape of the MarketConfig object, avoiding the need to label specific markets with specific provider names. I backed into this solution because our approach was beginning to feel cumbersome. the ticket i spun out for this is to review all providers and market types (lend/borrow/swap) and if we do want to continue to go with this labeling approach, we should commit fully and implement for all other market types, not just aave/morpho borrow markets.

its-everdred:
Merging as it's needed for the next PR, feel free to continue commenting and I can roll fixes into #490

PR #490, opened 2026-06-01
AaveBorrowProvider demo integration

its-everdred:
## Problem

Morpho supports lend + borrow, but Aave supports lend only — there is no Aave borrowing.

## Solution

AaveBorrowProvider — demo integration (#468). Stacked on #489.

aave-borrow

- Real Aave borrow on OP Sepolia, mirrored as `USDC_DEMO` on Base Sepolia (mint on borrow, remove on repay) — backend and in-browser (Turnkey) wallet paths; silent, best-effort.
- Lend = collateral: Morpho lend shares auto-pledged via reconciliation, so you borrow against the full lend; borrow draws against existing collateral (never auto-pledges).
- Health from one (stub) price source so borrowing always raises LTV; repay gated on held debt-asset balance with a "get more" prompt.
- `aave-v3` threaded through frontend marketId helpers; Aave market + `USDC` in demo config/allowlists.
- Fully-repaid positions hidden; Repay disabled at zero debt; withdraw Max no longer false-liquidates.
- Sticky nav + activity sidebar; landing page: Aave/Morpho borrow logos, "OP Stack chains", dynamic year, wider logo spacing.

## Demo magic

Two protocol designs need different sleights to present one clean "borrow against your lend" flow:

- **Morpho is isolated markets → emulated with two markets.** "Lend, then borrow against it" is stitched from a lend **vault** (deposit `USDC_DEMO` → vault shares) plus a separate Morpho Blue **borrow market** whose collateral token *is* those vault shares. Problem: the lent shares sit in the vault, not the borrow market, so they aren't usable as collateral. Fix: on the borrow tab we auto-pledge the vault shares to the borrow market (reconciliation), so you borrow against your full lend, and "Lent Balance" sums in-vault + pledged shares.
- **Aave is one shared pool → emulated with two USDCs.** The demo borrows **real USDC** against real ETH on OP Sepolia, Aave doesn't support permissionless asset markets additions. The demo economy is denominated in **`USDC_DEMO`** (a mintable mock on Base Sepolia). Problem: borrowing real USDC gives the user nothing spendable in the demo. Fix: mirror the two — mint `USDC_DEMO` 1:1 after the real borrow and remove it (transfer to a dead sink) after the repay — so the user holds/spends `USDC_DEMO` while the real debt lives on Aave.
- **Prices are stubbed.** Health and LTV are computed from a single local stub price source rather than each protocol's live oracle. This keeps borrowing's effect on the health bar predictable and the two demo economies comparable, and deliberately avoids wiring real per-protocol price feeds in this PR. A shared price provider can replace the stub later.


packages/cli/src/demo/markets.ts
```diff
@@ -52,8 +82,6 @@ export const MorphoUSDCBorrowOPDemo: BorrowMarketConfig = {
   name: 'Demo dUSDC / OP',
   collateralAsset: USDC_DEMO,
   borrowAsset: OP_DEMO,
-  borrowProvider: 'morpho',
-  lendProvider: 'morpho',
```
its-everdred, line 56: these were unnecessary.

jefr90 (approved):
lgtm 🔥

PR #500, opened 2026-06-18
Fix Velodrome universal router recipient routing

### LINKED ISSUE #444: Velodrome universal router silently ignores user-supplied recipient
The user-supplied `recipient` to `wallet.swap.execute({ ..., recipient })` is silently ignored on universal-router chains (today: Base Sepolia). Tokens-out go to `wallet.address` regardless. No revert, no warning — the swap completes "successfully" into the wrong account.

Pre-existing bug, independent of #438 / #443. Surfaced while reading the encoder in that PR.

## The mechanic

`UNIVERSAL_ROUTER_MSG_SENDER` is a magic-address constant baked into the Universal Router. When the router sees this sentinel in a command's `recipient` slot, it resolves it to `msg.sender` of the `execute()` call at runtime. The SDK encoder hardcodes the sentinel for both V2 and CL paths:

- [`encoding/routers/v2.ts:220`](https://github.com/ethereum-optimism/actions/blob/main/packages/sdk/src/actions/swap/providers/velodrome/encoding/routers/v2.ts#L220):

  ```ts
  UNIVERSAL_ROUTER_MSG_SENDER, // recipient = msg.sender (Universal Router sentinel)
  ```

- [`encoding/routers/cl.ts:143`](https://github.com/ethereum-optimism/actions/blob/main/packages/sdk/src/actions/swap/providers/velodrome/encoding/routers/cl.ts#L143): same.

Output tokens always go back to whoever called `execute()` — the wallet itself.

## The mismatch

The SDK's public API exposes a `recipient` param. `WalletSwapNamespace` resolves it (including ENS), the base provider plumbs it through `params.recipient` and into the encoders. **v2 and leaf routers honor it** ([v2.ts:253, 261, 268](https://github.com/ethereum-optimism/actions/blob/main/packages/sdk/src/actions/swap/providers/velodrome/encoding/routers/v2.ts#L253) — passes `recipient` directly as the swap's `to:` arg). The universal-router encoder discards it and substitutes the sentinel.

End state on a universal-router chain: tokens land in `wallet.address`, not the user-supplied recipient.

## Practical impact

| Caller pattern | v2/leaf chain | Universal-router chain |
|---|---|---|
| No `recipient` (defaults to `wallet.address`) | ✓ correct | ✓ correct (sentinel resolves to same address) |
| `recipient: someOtherAddress` | ✓ tokens go to that address | ✗ silently goes to `wallet.address` |

It's invisible in default usage and that's almost certainly why it was never noticed: the CLI doesn't expose `--recipient`, the demo backend uses the default, #438's repro doesn't touch it. **The bug only manifests when an integrator explicitly passes a non-default recipient on a universal-router chain.**

## Why the "recipient changed" guardrail doesn't catch it

The base `SwapProvider` has a re-encode path that fires when the recipient changes between quote-time and execute-time ([SwapProvider.ts:436](https://github.com/ethereum-optimism/actions/blob/main/packages/sdk/src/actions/swap/providers/velodrome/encoding/routers/v2.ts)):

```ts
if (quote.recipient !== quote.quotedRecipient) {
  const freshQuote = await this._getQuote({...})
  return this.buildSwapTransactions({ ...freshQuote, recipient: quote.recipient })
}
```

This guardrail exists *precisely because* recipient is normally baked into the calldata. But on universal-router chains the calldata doesn't depend on `recipient` — it's always the sentinel. So the freshly-quoted calldata is byte-identical to the stale one, the safeguard does nothing, and there's no warning that the user's intent has been silently dropped.

## Severity

- **Today**: latent. Base Sepolia is the only universal-router chain configured. It's a testnet and no caller in this repo passes a non-default recipient. Worst case is a confused integrator who silently sees tokens stay in the wallet.
- **When universal router rolls to mainnet** — Velodrome already has it deployed at `0x6Cb442acF35158D5eDa88fe602221b67B400Be3E` on Base mainnet, [`addresses.ts`](https://github.com/ethereum-optimism/actions/blob/main/packages/sdk/src/actions/swap/providers/velodrome/addresses.ts) just doesn't target it yet — this becomes a real fund-misdirection footgun.
- **ZeroDev session-key flows (post-#403)**: the natural pattern is "smart wallet executes the swap, output goes to a separate user EOA." Currently broken on universal-router chains.

## Recommended fix

Pass `params.recipient` literally instead of the sentinel. The Universal Router accepts any address in the `recipient` field; the sentinel is purely an opt-in convenience.

Concrete changes:

- `encoding/routers/v2.ts:220`: replace `UNIVERSAL_ROUTER_MSG_SENDER` with `params.recipient`.
- `encoding/routers/cl.ts:143`: replace `UNIVERSAL_ROUTER_MSG_SENDER` with `params.recipient`.
- Drop the `UNIVERSAL_ROUTER_MSG_SENDER` re-export if it becomes dead code (check `helpers.ts`).

`params.recipient` is already required on the encoder's `EncodeSwapParams` / `EncodeCLSwapParams` interfaces and resolved upstream in `_getQuote` via `resolveQuoteDefaults` (defaults to `quote.recipient ?? UNIVERSAL_ROUTER_MSG_SENDER` — that fallback should also drop the sentinel and resolve to `wallet.address` higher up, since `WalletSwapNamespace` already injects it).

### Related: split signer address from output recipient in `SwapQuote`

Surfaced during code review of #443. The `SwapQuote.recipient` field is doing double duty:

- Documented as **"Wallet address that will execute this quote"** ([`base.ts:242-246`](https://github.com/ethereum-optimism/actions/blob/main/packages/sdk/src/types/swap/base.ts#L242)) — i.e. the signer
- Used downstream as the **output recipient** (passed to encoders) **and** as the **allowance owner** (passed as `owner` to `buildApprovalTxIfNeeded` from `VelodromeSwapProvider._buildApprovals`)

These collapse to the same address today only because the SDK always defaults recipient to `wallet.address`. Once this issue's fix lets callers supply a custom recipient, the allowance owner must remain the **signer** (the address that calls `execute()` and holds the input tokens) — not the recipient (the address that receives output).

The fix above ("replace sentinel with `params.recipient`") implicitly assumes the encoder receives a true output-recipient. If the upstream pipeline still sets `recipient = wallet.address` even when the caller passed a custom value, the encoder fix is a no-op. So #444 needs to:

1. Add a separate `walletAddress` (or `signerAddress`) field to `SwapQuote` / `NormalizedSwapQuote`, set by `WalletSwapNamespace` to the actual signer.
2. Update `VelodromeSwapProvider._buildApprovals` (currently passes `owner: quote.recipient` to `buildApprovalTxIfNeeded`) to use the new signer field for the allowance owner.
3. Stop defaulting `recipient` to `wallet.address` upstream once the two fields are separate; let the user-supplied value flow through unchanged.

Without this split, the encoder fix risks emitting an `approve` against the wrong owner address — either an unnecessary approve when the signer already has allowance, or a missed approve that causes `execute()` to revert because the recipient (not the signer) was checked for allowance.

## Test plan

- **Encoding regression** (`encoding.v2.test.ts`, `encoding.cl.test.ts`): decode the inner `V2_SWAP_EXACT_IN` / `V3_SWAP_EXACT_IN` input from the universal-router `execute()` calldata and assert that the `recipient` field equals the supplied recipient address (not the sentinel). Test both:
  - default case (`recipient = wallet.address`) — should encode `wallet.address`, not the sentinel
  - custom case (`recipient = someOther`) — should encode `someOther`
- **Provider regression** (`VelodromeSwapProvider.test.ts`): on Base Sepolia, calling `execute({ recipient: someOther })` produces calldata whose decoded inner recipient equals `someOther`. This is the unit-level repro that fails before the fix.
- **Allowance-owner regression**: with a custom `recipient` on Base Sepolia, the allowance check in `_buildApprovals` reads from the signer's address, not the recipient's. Mock `allowance(signer, router) = MAX` and `allowance(recipient, router) = 0` and assert no `approve` is emitted.
- The existing v2/leaf recipient assertions in `VelodromeSwapProvider.test.ts` and `routing.test.ts` continue to pass unchanged.
- **Manual end-to-end** before merge: a swap on Base Sepolia with an explicit `recipient` argument lands the tokens at the recipient address, not at the wallet.

## Out of scope

- Whether other providers (Uniswap V4) have the same shape. Worth a separate audit but tracked under different code.
- Adding `--recipient` support to the CLI. Orthogonal feature.

## See also

- #438, #443 — the universal-router `payerIsUser` fix where this was surfaced
- #396 — wallet-layer refactor (forward-compatible with this fix)

its-everdred:
Closes #444

# Problem

This PR fixes two related recipient defaults:
First, wallet-bound swaps need a safe generic fallback: when a caller does not pass `recipient` but the SDK knows the executing `walletAddress`, the quote should send output back to that wallet while approvals and allowance checks also use that wallet as the token owner. That fallback is safe because the wallet namespace injects the actual executing wallet address, explicit recipients still override it, and it avoids sending output to an unrelated placeholder.

Second, `UNIVERSAL_ROUTER_MSG_SENDER` is not a generic recipient default. It is a Universal Router calldata sentinel meaning "use `msg.sender`", so it only makes sense in providers and router branches that actually encode Universal Router commands. Keeping that sentinel in `SwapProvider` made a router-specific convention look valid for all current and future swap providers, including Velodrome's v2 and leaf routers that expect concrete recipient addresses.

The concrete failure path was: configure Velodrome on a v2 or leaf-router chain, request a swap quote without an explicit recipient, then execute that quote. The old abstract default could populate `quote.recipient` with `UNIVERSAL_ROUTER_MSG_SENDER`; those routers treat that as the literal `0x0000000000000000000000000000000000000001` recipient, not as `msg.sender`, so successful swaps could route output away from the executing wallet.

# Solution

- Preserve the caller-requested swap recipient in encoded calldata.
- Keep `walletAddress` separate for signer and allowance ownership.
- Keep wallet-address recipient fallback in shared defaults.
- Scope `UNIVERSAL_ROUTER_MSG_SENDER` fallback to Uniswap and Velodrome's Universal Router path only.
- Require an explicit recipient for non-Universal Router Velodrome quotes.
- Add regressions for wallet-bound quotes, custom recipients, and provider-specific recipient defaults.

# Validation

- `pnpm --filter @eth-optimism/actions-sdk test -- SwapProvider.test.ts UniswapSwapProvider.test.ts VelodromeSwapProvider.test.ts WalletSwapNamespace.spec.ts`
- `pnpm --filter @eth-optimism/actions-sdk typecheck`
- `pnpm --filter @eth-optimism/actions-sdk lint` (passes with existing warning baseline)


packages/sdk/src/actions/swap/core/SwapProvider.ts
```diff
@@ -271,15 +271,18 @@ export abstract class SwapProvider<
...
+   * amountInRaw, and current timestamp
    */
   protected resolveQuoteDefaults(params: SwapQuoteParamsResolved) {
     const slippage = params.slippage ?? this.defaultSlippage
     const now = Math.floor(Date.now() / 1000)
     const deadline = params.deadline ?? now + this.quoteExpirationSeconds
-    const recipient = params.recipient ?? UNIVERSAL_ROUTER_MSG_SENDER
+    const recipient =
+      params.recipient ?? params.walletAddress ?? UNIVERSAL_ROUTER_MSG_SENDER
+    const walletAddress = params.walletAddress
```
its-everdred: Implemented in 697069e2: the base provider no longer applies the sentinel, Uniswap owns its Universal Router fallback, and Velodrome only uses it for Universal Router chains.

packages/sdk/src/actions/swap/core/SwapProvider.ts
```diff
@@ -270,16 +269,20 @@ export abstract class SwapProvider<
...
-   * @returns Resolved slippage, deadline, recipient, amountInRaw, and current timestamp
+   * @returns Resolved slippage, deadline, optional recipient and wallet address,
+   * amountInRaw, and current timestamp
    */
   protected resolveQuoteDefaults(params: SwapQuoteParamsResolved) {
     const slippage = params.slippage ?? this.defaultSlippage
     const now = Math.floor(Date.now() / 1000)
     const deadline = params.deadline ?? now + this.quoteExpirationSeconds
-    const recipient = params.recipient ?? UNIVERSAL_ROUTER_MSG_SENDER
+    const recipient = params.recipient ?? params.walletAddress
```
its-everdred: This line contains both fixes

packages/sdk/src/actions/swap/core/SwapProvider.ts
```diff
@@ -461,6 +464,15 @@ export abstract class SwapProvider<
     }
   }
 
+  /**
+   * Resolve the wallet that owns input tokens for a quote.
+   * @param quote - Swap quote being executed
+   * @returns Wallet address for allowance checks
+   */
+  protected resolveQuoteWalletAddress(quote: SwapQuote): Address {
+    return quote.walletAddress ?? quote.recipient
```
its-everdred: i dont like that we have this walletAddress -> recipient fallback here and then above its:
params.recipient -> params.walletAddress. 

is it intended these differ? can these be consolidated in some way?

its-everdred: Yes, they are intentionally opposite directions today, but I agree the duplication makes that hard to reason about. I’ll consolidate this into named helper(s) so quote creation defaults `recipient` from the executing wallet, while quote execution only falls back from `walletAddress` to `recipient` for quote compatibility.

its-everdred: Implemented in e2af2f1a: shared `recipient ?? walletAddress` fallback now lives in `resolveDefaultRecipient`, while quote execution keeps the separate compatibility fallback for allowance ownership. Focused tests, typecheck, and lint pass.

packages/sdk/src/actions/swap/core/SwapProvider.ts
```diff
@@ -530,6 +541,23 @@ export abstract class SwapProvider<
...
+  private resolveDefaultRecipient(params: {
+    recipient?: Address
+    walletAddress: Address
+  }): Address
+
+  private resolveDefaultRecipient(params: {
+    recipient?: Address
+    walletAddress?: Address
+  }): Address | undefined
+
```
its-everdred: question: why do we need these two overloads?

its-everdred: They preserve type precision for two call sites: quote defaults may not have `walletAddress`, so the helper can return `undefined`, while execute params always have `walletAddress`, so `ResolvedSwapParams.recipient` stays an `Address`. That said, the overloads are too much ceremony for this fallback; I’ll simplify this so the execute path remains type-safe without exposing overloads.

its-everdred: Simplified in 2b5d5168: removed the overloads and kept a single optional helper; the execute path now explicitly falls back to its required `walletAddress` to satisfy `ResolvedSwapParams`. Focused swap tests, typecheck, and lint pass.

packages/sdk/src/actions/swap/namespaces/WalletSwapNamespace.ts
```diff
@@ -34,37 +34,43 @@ export class WalletSwapNamespace extends BaseSwapNamespace {
...
-   * Get a swap quote with the wallet address as recipient.
-   * Ensures calldata is encoded for the real wallet, not a placeholder.
+   * Get a swap quote bound to the wallet as executor.
+   * Defaults the output recipient to the wallet address when omitted.
    */
   override async getQuote(params: SwapQuoteParams): Promise<SwapQuote> {
-    return super.getQuote({
+    const quote = await super.getQuote({
       ...params,
       recipient: params.recipient ?? this.wallet.address,
```
its-everdred: https://github.com/ethereum-optimism/actions/pull/500/changes#r3484077479 this conversation is about where we codify this fallback within the context of the swap provider one. Is it duplicative in a bad way that both locations codify this separately? I guess the actions swap namespace may have another instance of it? I'm wondering if it makes sense to consolidate all instances of this logic to a single utility that gets called in all places. Alternatively, I'm wondering if it makes sense to have the fallback occur once per request, either at the namespace level or at the abstract provider level. Thoughts?

its-everdred: Agreed. I think request-level fallback should be centralized once as `recipient ?? walletAddress` and used by both `WalletSwapNamespace` and `SwapProvider`; the quote `walletAddress ?? recipient` fallback should stay separate because it is execution compatibility for allowance ownership, not request defaulting. I’ll implement the small shared helper and keep those two concepts distinct.

packages/sdk/src/actions/swap/namespaces/WalletSwapNamespace.ts
```diff
@@ -34,37 +34,43 @@ export class WalletSwapNamespace extends BaseSwapNamespace {
...
-    return super.getQuotes({
+    const quotes = await super.getQuotes({
       ...params,
       recipient: params.recipient ?? this.wallet.address,
+      walletAddress: this.wallet.address,
     })
+    return quotes.map((quote) => ({
+      ...quote,
+      walletAddress: this.wallet.address,
+    }))
```
its-everdred, line 71: Can you explain why you made this change? I'm not questioning it. I just don't understand.

packages/sdk/src/actions/swap/namespaces/WalletSwapNamespace.ts
```diff
@@ -86,14 +92,14 @@ export class WalletSwapNamespace extends BaseSwapNamespace {
...
    * Validate that a pre-built quote is bound to this wallet. Throws when the
-   * quote's recipient differs from `wallet.address`; silently swapping
-   * recipients would route output tokens to the wrong address on routers that
-   * encode the recipient directly into calldata (e.g. Velodrome v2/leaf).
+   * quote's wallet address differs from `wallet.address`; silently executing
+   * with a different wallet would check allowances for the wrong account.
    */
   private requireQuoteForThisWallet(quote: SwapQuote): SwapQuote {
-    if (!isAddressEqual(quote.recipient, this.wallet.address)) {
+    const quoteWalletAddress = quote.walletAddress ?? quote.recipient
```
its-everdred: Here's another instance of this fallback logic that needs to be consolidated.

packages/sdk/src/actions/swap/providers/uniswap/UniswapSwapProvider.ts
```diff
@@ -151,12 +156,13 @@ export class UniswapSwapProvider extends SwapProvider<UniswapSwapProviderConfig>
           }
     const amountInMaxRaw =
       'amountInMaxRaw' in swapAmounts ? swapAmounts.amountInMaxRaw : undefined
+    const quoteRecipient = recipient ?? UNIVERSAL_ROUTER_MSG_SENDER
```
its-everdred: This is another important fallback piece of logic that feels like it deserves its own utility function within the Uniswap provider.

packages/sdk/src/types/swap/base.ts
```diff
@@ -91,7 +91,10 @@ export interface WalletSwapParams {
   slippage?: number
   /** Transaction deadline as Unix timestamp. Defaults to now + 1 minute. */
   deadline?: number
-  /** Recipient address or ENS name (e.g. "vitalik.eth"). Defaults to wallet address. */
+  /**
+   * Recipient address or ENS name (e.g. "vitalik.eth").
+   * Wallet namespaces default this to wallet address.
+   */
   recipient?: Address | EnsName
```
its-everdred, line 98: This line makes me feel like we should do a quick review spike on all instances of recipient that this touches, to ensure that we actually support ENS name correctly and that passing that doesn't trickle into some field that we didn't anticipate, like wallet address.

packages/sdk/src/actions/swap/namespaces/WalletSwapNamespace.ts
its-everdred: That mirrors `getQuote`: `getQuotes` is also wallet-bound, so every provider quote needs both the executing wallet address and the default output recipient when the caller omits `recipient`. The helper now makes that the same request-level rule instead of leaving this as another local fallback.

packages/sdk/src/actions/swap/namespaces/WalletSwapNamespace.ts
its-everdred: Agreed this should be named, but it should stay separate from request recipient defaulting. This fallback is quote-compatibility logic for older/prebuilt quotes: if `walletAddress` is missing, use the quote recipient as the wallet binding check; I’ll extract that into a focused helper so the direction is explicit.

packages/sdk/src/actions/swap/providers/uniswap/UniswapSwapProvider.ts
its-everdred: Agreed. This is provider-specific Universal Router fallback logic, so I’ll extract it inside the Uniswap provider rather than leaving the sentinel fallback inline.

packages/sdk/src/types/swap/base.ts
its-everdred: Agreed. Before further changes I’ll audit the touched recipient paths for ENS handling and make sure `EnsName` only flows through namespace resolution, while provider execution, wallet ownership, and allowance checks receive concrete `Address` values.

packages/sdk/src/actions/swap/namespaces/WalletSwapNamespace.ts
its-everdred: Implemented in 335fdbef: request-level `recipient ?? walletAddress` now lives in `resolveSwapRequestRecipient`, and the quote wallet-binding fallback is separate as `resolveSwapQuoteWalletAddress`. Focused swap tests, typecheck, and lint pass.

packages/sdk/src/actions/swap/namespaces/WalletSwapNamespace.ts
its-everdred: Implemented in 335fdbef: this now uses `resolveSwapQuoteWalletAddress`, which names the compatibility fallback separately from request recipient defaulting. Focused swap tests, typecheck, and lint pass.

packages/sdk/src/actions/swap/providers/uniswap/UniswapSwapProvider.ts
its-everdred: Implemented in 335fdbef: Uniswap now uses `resolveUniswapUniversalRouterRecipient`, keeping the Universal Router sentinel fallback provider-local and explicit. Focused swap tests, typecheck, and lint pass.

packages/sdk/src/types/swap/base.ts
its-everdred: Audit complete in 335fdbef: request recipient fallback remains generic enough for `EnsName`, but namespace quote/execute paths resolve ENS before providers receive params. The quote wallet helper only accepts `Address`, so ENS cannot flow into wallet ownership or allowance checks. Focused swap tests, typecheck, and lint pass.

packages/sdk/src/actions/swap/recipients.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Resolve request-level swap recipient defaults.
+ * @param recipient - Explicit output recipient, when provided
+ * @param walletAddress - Wallet fallback for wallet-bound requests
+ * @returns Explicit recipient, or wallet fallback when omitted
+ */
+export function resolveSwapRequestRecipient<
+  TRecipient,
+  TWalletAddress extends TRecipient | undefined,
+>(
+  recipient: TRecipient | undefined,
```
tremarkley, line 15: is the main benefit of allowing recipient to be undefined, so that someone using the sdk does not have to add the wallet address themselves? Open to being wrong here, but might it be better to require this so they have to be explicit about where the swap is going and also removes the need of us having to support a sentinel value?

packages/sdk/src/actions/swap/recipients.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Resolve the wallet address that owns input tokens for a quote.
+ * @param quote - Swap quote recipient and optional wallet binding
+ * @returns Wallet address for wallet validation and allowance checks
+ */
+export function resolveSwapQuoteWalletAddress(quote: {
+  walletAddress?: Address
+  recipient: Address
+}): Address {
+  return quote.walletAddress ?? quote.recipient
```
tremarkley: so if the quote doesnt have a `walletAddress` on it we redirect to a `recipient`? the comment above says "resolve the wallet address that owns input tokens for a quote", so in this case if `walletAddress` is undefined, then `recipient` is the owner of the tokens?

tremarkley (approved):
just a couple questions, but looks good

packages/sdk/src/actions/swap/recipients.ts
its-everdred: Yeah this is bad. Currently, if a dev gets a generic quote through `actions.swap.getQuote(...)`, we have fallback logic that can carry it into `wallet.swap.execute(...)` by treating the quote’s output recipient as the executing wallet. It's trying to be too clever and bakes in bad assumptions, so I'm updating this to require quote.walletAddress for wallet-bound execution and ask devs to re-quote through `wallet.swap.getQuote(...)` instead.

packages/sdk/src/actions/swap/recipients.ts
its-everdred: Yea that's the main benefit.
Since Actions owns the context of both the wallet and the swap, I'm inclined to consider the user's wallet address a safe default, especially since the swap function is invoked directly off of the wallet object. `wallet.swap.execute({walletAddress: wallet.address, ...})` feels cumbersome to require.

I'd also describe the ergonomics and simplicity of the actions abstractions to be a value prop. 

For comparison, uniswap has lower level sdk. it wrap's the swap contract functions closely, requiring a wallet address, and then you pass the call data to the wallet directly ex:

```
  const calldata = SwapRouter.swapCallParameters(trade, { recipient: wallet.address, slippageTolerance, deadline })
  await walletClient.sendTransaction({ to: SWAP_ROUTER, data: calldata.calldata, value: calldata.value })
```

But, velodrome's Sugar SDK (possible reference to syntactic sugar?) is a higher level, letting you set the wallet address at a higher level context and then the swap function just passes along that address as the default.
python:
```
  op = AsyncOPChain(signer_address=wallet.address)
  txs = await op.swap(from_token=VELO, to_token=ETH, amount=amount)
  await wallet.sign_and_send_all(txs)
```

tremarkley: sounds good. 

On a side note, as I was looking further into this. Claude identified a potential related bug to this. I havent fully vetted it but wanted to pass it along so you are aware and can determine whether it is an actual bug and what the appropriate fix is. Here is a copy pasta of the Claude output:

But pulling on the "do we even need the sentinel" thread led me somewhere I think matters more than the optional-vs-required question: **on the Uniswap path, the `recipient` we resolve here never makes it into the calldata.**

`encodeUniversalRouterSwap` builds its V4 action sequence as `SWAP_EXACT_IN_SINGLE → SETTLE_ALL → TAKE_ALL`. The `recipient` arg isn't a parameter of any of those actions — and `TAKE_ALL` hardcodes the destination to the caller ([V4Router.sol](https://github.com/Uniswap/v4-periphery/blob/main/src/V4Router.sol)):

```solidity
else if (action == Actions.TAKE_ALL) {
    (Currency currency, uint256 minAmount) = params.decodeCurrencyAndUint256();
    uint256 amount = _getFullCredit(currency);
    if (amount < minAmount) revert V4TooLittleReceived(minAmount, amount);
    _take(currency, msgSender(), amount);   // always msg.sender, no recipient decoded
}
```

Only `TAKE` (`0x0e`) decodes a recipient and runs it through `_mapRecipient` — the function that makes `address(1)` resolve to `msg.sender` and `address(2)` to the router.

**Impact:**
- In the wallet-bound flow it's invisible: `recipient` defaults to `walletAddress` == `msg.sender`, so output lands correctly anyway.
- But if a caller passes an explicit `recipient` that isn't their own wallet (swap → send output elsewhere), the quote's `recipient` metadata says one address while the calldata pays `msg.sender`. The recipient is silently dropped.
- And to your original point: the `UNIVERSAL_ROUTER_MSG_SENDER` sentinel is currently a **no-op** on the Uniswap path — `TAKE_ALL` already routes to `msg.sender` no matter what we pass.

Velodrome is unaffected — its universal/v2/leaf encoders all thread `recipient` into the command/router args (it's the first field of `V2_SWAP_EXACT_IN_INPUT_PARAMS`), so custom recipients are honored there.

**Suggested fix** — swap the trailing `TAKE_ALL` for `TAKE`, passing the resolved recipient and the full output delta:
- `TAKE(currency = tokenOut, recipient = quoteRecipient, amount = OPEN_DELTA)`, where `OPEN_DELTA = 0` resolves via `_mapTakeAmount` → `_getFullCredit(currency)` to the entire swap output ([DeltaResolver.sol](https://github.com/Uniswap/v4-periphery/blob/main/src/base/DeltaResolver.sol)).
- `_mapRecipient` still maps the `address(1)` sentinel → `msg.sender`, so the wallet-default case keeps working unchanged (and the sentinel becomes meaningful again instead of dead).
- Slippage stays enforced by `amountOutMinimum` on `SWAP_EXACT_IN_SINGLE`, so dropping `TAKE_ALL`'s `minAmount` check loses no protection.

So to close out the original question: I don't think we need to require `recipient` globally — the wallet default is fine. The sentinel does earn its keep, but only on the wallet-less `actions.swap.getQuote` path (no wallet to default to); everywhere wallet-bound it's redundant with the `walletAddress` default, so scoping it per-provider is the right call. The thing I'd want before merge is the Uniswap `TAKE_ALL → TAKE` fix plus a test asserting a custom recipient actually lands on-chain for **both** providers — this folds directly into the ENS-recipient audit on `base.ts:98`.

PR #501, opened 2026-06-18
CLI: ENS Support

### LINKED ISSUE #453: CLI: ENS namespace (resolve, reverse, info)
Parent: #407.

The Actions CLI scaffolding (#408 / #421) and the lend namespace (#409 / #423) cover `actions.lend.*` and `actions.wallet.*` write paths, but the SDK's `actions.ens.*` namespace currently has no CLI surface.

## SDK surface to mirror

`EnsNamespace` (`packages/sdk/src/services/nameservices/ens/EnsNamespace.ts`) exposes:

- `getAddress(input: Address | EnsName): Promise<Address>` — name → address (or echoes back when given an address).
- `getName(address: Address): Promise<EnsName | null>` — reverse-resolve.
- `getInfo(input: Address | EnsName): Promise<EnsInfo>` — combined `{ name, address }`.

## Proposed CLI surface

Read-only, no `PRIVATE_KEY` required:

- `actions ens resolve <name>` → emits `{ name, address }` (where address is derived).
- `actions ens reverse <address>` → emits `{ address, name }` (or `name: null` when no reverse record).
- `actions ens info <input>` → emits `EnsInfo` directly (the SDK shape).

## Notes

- **Mainnet requirement:** `EnsNamespace` requires Ethereum mainnet (chain ID 1) in the chain config. The CLI's demo config currently doesn't include mainnet — it ships with Base Sepolia + Optimism Sepolia. The handler should surface a clean `CliError('config', ...)` when mainnet isn't configured rather than relying on the SDK's `EnsNotConfiguredError` to bubble through. (After PR #423 lands, the typed `ActionsError` mapping in `output/errors.ts` does this automatically.)
- **Public RPC fallback:** the SDK falls back to a public mainnet RPC when mainnet isn't in the chain config. Decide whether the CLI should accept this fallback or insist on operator-configured RPC for ENS reads (consistent with the "RPC trust" doc in `SKILL.md`).
- **ENS name validation:** the SDK has `isEnsName(input)` for shape-checking. Use it at the resolver layer so unknown inputs surface as `validation` rather than `network`.

## Why this is its own issue

Originally surfaced as a gap during the #423 review (`/compound-engineering:workflows:review`). #421 stays scoped to scaffolding and #423 to lend; ENS is small and self-contained but logically unrelated to either.

## Acceptance criteria

- [ ] `actions ens resolve <name>` and `actions ens reverse <address>` work end-to-end against mainnet.
- [ ] `actions ens info <input>` emits the SDK `EnsInfo` shape verbatim.
- [ ] Missing-mainnet config → `CliError('config')` exit 3.
- [ ] Invalid ENS name shape → `CliError('validation')` exit 2.
- [ ] `SKILL.md` documents the new commands with NL→command examples.

its-everdred:
Closes #453

Related comment: https://github.com/ethereum-optimism/actions/pull/501#issuecomment-4918028000

# Problem
The CLI did not expose the SDK's ENS namespace, so users and agents had no shell path for resolving names, reverse names, or ENS profile records. ENS reads also depend on Ethereum mainnet, which is awkward when the CLI app chains are configured for other networks and the SDK may need to fall back to a public RPC. Borrow position lookups should stay aligned with the SDK's address-only API instead of teaching the CLI to resolve extra wallet identifiers.

# Solution
Adds read-only ENS commands and keeps borrow position as an address-only wallet lookup.

- Adds `actions ens address`, `actions ens name`, and `actions ens info` with typed validation and sanitized human-readable output.
- Keeps `actions borrow position` on `--wallet <address>`, matching the SDK's `walletAddress` parameter without custom ENS resolution in the CLI.
- Allows SDK L1 chain construction for mainnet and sepolia via viem chain definitions, so configured mainnet RPCs can back ENS reads.
- Warns once when CLI ENS reads use the SDK public mainnet fallback instead of an operator-provided `MAINNET_RPC_URL`.
- Routes ENS RPC failures into retryable CLI network errors and shares output sanitization across success and error paths.


its-everdred:
i dont like this name: 
```
actions ens resolve <name>
actions ens reverse <address>
```

instead lets do 
```
 actions ens address <name>
actions ens name <address>
```

packages/cli/src/output/printOutput.ts
```diff
@@ -144,6 +159,41 @@ function formatAddress(doc: Printers['address']): void {
   writeLine(doc.address)
 }
 
+// ENS names and profile text records are arbitrary on-chain strings set by
+// whoever owns the name. Rendered verbatim in a terminal they can carry ANSI /
+// OSC escape sequences that rewrite the screen, the window title, or inject
+// clickable phishing hyperlinks. Strip C0/C1 control bytes (keeping ordinary
+// printable text) before writing the human-readable output. The `--json` path
+// is unaffected: JSON.stringify already escapes control characters.
+const CONTROL_CHARS = /[\u0000-\u001F\u007F-\u009F]/g
```
its-everdred: This comment should tell me what these chars are, and why these are a special case

packages/cli/src/resolvers/ens.ts
```diff
@@ -0,0 +1,90 @@
...
+export function requireMainnet(config: NodeActionsConfig<never>): void {
+  if (!config.chains.some((chain) => chain.chainId === mainnet.id)) {
+    throw new CliError(
+      'config',
+      'ENS operations require Ethereum mainnet (chain ID 1). ' +
+        'Set MAINNET_RPC_URL to a trusted endpoint to enable ENS reads.',
+      { chainId: mainnet.id },
+    )
+  }
+}
```
its-everdred: I'm not sure if I agree that the CLI should be opinionated in this way. if the SDK calls back to a public RPC, the CLI should just be a thin wrapper. we can add a warning here instead in this case, but dont fail

packages/cli/src/resolvers/ens.ts
```diff
@@ -0,0 +1,90 @@
...
+export function requireAddress(input: string): Address {
+  if (!isAddress(input)) {
+    throw new CliError(
+      'validation',
+      `Invalid address: "${input}" (expected a 0x-prefixed 20-byte address)`,
+      { input },
+    )
+  }
+  return getAddress(input)
+}
```
its-everdred: This validation is probably not ENS specific and may already exist elsewhere. if not, move to a util and import

packages/cli/src/resolvers/ens.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export function requireEnsNameOrAddress(input: string): Address | EnsName {
+  if (isAddress(input)) return getAddress(input)
+  if (isEnsName(input)) return input
+  throw new CliError(
+    'validation',
+    `Invalid input: "${input}" (expected an ENS name or a 0x-prefixed address)`,
+    { input },
+  )
+}
```
its-everdred: Is this used anywhere? if not delete

packages/cli/SKILL.md
```diff
@@ -311,11 +318,49 @@ NL -> command examples:
...
+markets are testnet-only, so this affects reads (balances), not write targets.
+
+NL -> command examples:
+
+- "what address is vitalik.eth" -> `actions --json ens resolve vitalik.eth`
+- "resolve vitalik.eth" -> `actions --json ens resolve vitalik.eth`
+- "what's the ENS name for 0xd8dA...96045" -> `actions --json ens reverse 0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
+- "show vitalik.eth's profile" -> `actions --json ens info vitalik.eth`
+- "get the ENS records for 0xd8dA...96045" -> `actions --json ens info 0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`
+
```
its-everdred, line 357: The skill should also include more complex asks where the ens lookup is a step within another action. ex: "swap 1 USDC for eth and send it to vitalik.eth". in this case, the agent should know to use the swap recipient field. 

related: do the CLI address-related params allow passing of either address or ens names the way the sdk natively does? if not, it should

packages/sdk/src/services/ChainManager.ts
```diff
@@ -33,6 +33,33 @@ function pollingIntervalForChain(chainId: SupportedChainId): number {
...
+ * @param chainId - A {@link SupportedChainId} to resolve.
+ * @returns The viem {@link Chain}, or `undefined` when the id is unknown to
+ * both registries.
+ * @internal Not part of the public SDK surface; exported only so
+ * `MockChainManager` can mirror `ChainManager`'s resolution logic. No stability
+ * guarantee.
+ */
+export function viemChainFor(chainId: SupportedChainId): Chain | undefined {
+  return chainById[chainId] ?? L1_VIEM_CHAINS[chainId]
+}
```
its-everdred: I dont love this approach

packages/cli/src/output/printOutput.ts
its-everdred: > This comment should tell me what these chars are, and why these are a special case

Addressed: replaced the long block with a single-line comment that names C0/C1 as ANSI/OSC terminal controls and says why text ENS output strips them.

packages/cli/src/resolvers/ens.ts
its-everdred: > if the SDK calls back to a public RPC, the CLI should just be a thin wrapper. we can add a warning here instead in this case, but dont fail

Addressed: ENS commands now validate input, warn when `MAINNET_RPC_URL` is unset, and continue through the SDK public mainnet fallback instead of throwing config errors.

packages/cli/src/resolvers/ens.ts
its-everdred: > This validation is probably not ENS specific and may already exist elsewhere. if not, move to a util and import

Addressed: moved generic checksum-address validation into `utils/addresses.ts` and imported it from ENS reverse lookup and borrow `--wallet`.

packages/cli/src/resolvers/ens.ts
its-everdred: > Is this used anywhere? if not delete

Addressed: kept it because it is used by `ens info` and now by `borrow position --wallet <address|ens>` for shared address-or-ENS validation.

packages/cli/SKILL.md
its-everdred: > The skill should also include more complex asks where the ens lookup is a step within another action.

Addressed: added swap-recipient examples using `--recipient vitalik.eth`, documented `--recipient <addr|ens>`, and updated borrow `--wallet` to accept either a 0x address or ENS name before resolving it for `getPosition`.

packages/sdk/src/services/ChainManager.ts
its-everdred: > I dont love this approach

Addressed: removed the local L1 fallback map and now resolves `ChainManager` chain definitions from the SDK supported-chain registry, which already includes Ethereum L1 chains.

its-everdred:
> `actions ens resolve <name>` / `actions ens reverse <address>` -> `actions ens address <name>` / `actions ens name <address>`

Addressed: renamed the public commands and associated files/functions/tests to `address` and `name`, including `runEnsAddress`, `runEnsName`, `formatEnsAddress`, and `formatEnsName`.

packages/cli/src/commands/__tests__/actionsBorrowPosition.test.ts
```diff
@@ -23,13 +25,20 @@ describe('runBorrowPosition (read-only, with --wallet)', () => {
...
-      actions: { borrow: { getPosition } } as never,
+      config: {
+        ...config,
+        chains: config.chains.filter((chain) => chain.chainId !== 1),
+      },
+      actions: { borrow: { getPosition }, ens: { getAddress } } as never,
     })
   }
 
+  const VITALIK = '0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045'
```
its-everdred: rename VITALIK_ADDRESS

its-everdred: > rename VITALIK_ADDRESS

Addressed: moved the repeated ENS fixtures into `@/__tests__/helpers/ens.js`, renamed them to `MOCK_*`, and updated the similar CLI tests to use the shared helper with lowercase derived inline where needed.

jefr90 (approved):
lgtm 🔥

PR #508, opened 2026-06-18
SDK + CLI: getPositions() to fetch all wallet lend positions in one call

### LINKED ISSUE #454: SDK + CLI: getPositions() to fetch all wallet lend positions in one call
The SDK's `wallet.lend.getPosition({ marketId })` requires a specific `marketId` and throws `MarketIdRequiredError` otherwise (`packages/sdk/src/actions/lend/namespaces/WalletLendNamespace.ts:52`). There is no API to fetch all of a wallet's open positions across configured markets/providers. The same gap exists at the read-only `actions.lend.*` namespace.

## Why

Every consumer that wants to render \"the wallet's portfolio\" has to:

1. Walk the configured market allowlist.
2. Call `getPosition({ marketId })` once per market in parallel.
3. Filter zero-balance results client-side.

That's `M × P` RPC calls (markets × providers) per page-load — small for the demo allowlist, ugly for any larger config. It also makes batched optimisations (multicall, cf. #452) impossible because each SDK call is single-market.

### Existing demo cost

`packages/demo/frontend/src/hooks/useLendProvider.ts:106-139` already fans out manually:

\`\`\`ts
const positionPromises = marketInfoList.map(async (market) =>
  operations.getPosition({ ... })
)
const positionResults = await Promise.all(positionPromises)
\`\`\`

Two calls' worth of activity log entries (`getMarket` + `getPosition`) hide N RPCs. Replacing this with a single SDK call simplifies the hook AND makes activity-log accounting honest.

The post-trade refresh in `useLendPosition.ts:76-97, 129-150` already invalidates **one** position cache key (`['position', address, chainId]`); that pattern stays as-is — single-market `getPosition` is still the right tool when you know which market just changed.

## Proposed API

\`\`\`ts
// Read namespace (no wallet)
actions.lend.getPositions(walletAddress: Address, params?: GetPositionsParams): Promise<LendMarketPosition[]>

// Wallet namespace (uses wallet.address implicitly)
wallet.lend.getPositions(params?: GetPositionsParams): Promise<LendMarketPosition[]>

interface GetPositionsParams {
  /** Filter to one chain (mirrors GetLendMarketsParams). */
  chainId?: SupportedChainId
  /** Filter to one provider (\"morpho\" | \"aave\" etc.). */
  provider?: LendProviderName
  /** Drop zero-balance positions before returning. Default: false (return every market for parity with getMarkets). */
  nonZeroOnly?: boolean
}
\`\`\`

Returns the same `LendMarketPosition` shape the existing `getPosition` already uses — no new type.

## Implementation notes

- **Backed by multicall (#452 follow-up)**: every provider runs `getPosition` against viem `Multicall3` — one eth_call per chain instead of one per market. The savings stack with #452's balance optimization.
- **Per-call failure semantics**: `allowFailure: true` so a single bad market RPC doesn't poison the batch. Failures map to a typed `ActionsError` per market (or are dropped, depending on `nonZeroOnly`).
- **Provider aggregation**: `BaseLendNamespace.getPositions` runs `Promise.all` over configured providers; each provider's `_getPositions` walks its own allowlist. Symmetric to the existing `_getMarkets` pattern.

## CLI follow-up

Once the SDK ships `wallet.lend.getPositions`, add `actions wallet lend positions` (plural — the existing singular `position --market <name>` stays).

\`\`\`
actions wallet lend positions [--chain <name> | --chain-id <id>] [--non-zero-only]
\`\`\`

Output: `LendMarketPosition[]` verbatim, bigints stringified.

Until the SDK ships this, the CLI could implement the fan-out itself against `collectMarkets(config)`, but doing so client-side defeats the multicall optimization. Skip the CLI work until the SDK lands.

## Acceptance criteria

- [ ] `actions.lend.getPositions(address, params?)` and `wallet.lend.getPositions(params?)` exist on the SDK.
- [ ] Aggregates across all configured providers and markets.
- [ ] `LendMarketPosition[]` shape matches the existing `getPosition` return.
- [ ] Per-market failures don't poison the batch.
- [ ] Demo `useLendProvider.ts` consumes `getPositions` instead of the manual fan-out.
- [ ] CLI `actions wallet lend positions` ships.

Parent: #407.

its-everdred:
Closes #454

## Summary

Adds `getPositions()` to the SDK to fetch **all** of a wallet's lend positions in one call, replacing the per-market `getPosition` fan-out that every portfolio consumer had to write by hand. Wires it through the CLI and the demo.

### SDK
- `GetPositionsParams` (`chainId?`, `provider?`, `nonZeroOnly?`), exported from the public entry.
- `LendProvider.getPositions(walletAddress, params?)` — walks the provider's market allowlist and fetches each position with `Promise.allSettled`, so a single bad-market RPC can't poison the batch (rejected markets are dropped).
- `BaseLendNamespace.fetchPositions` — aggregates across configured providers (or one, via `provider`), flattens, then applies `nonZeroOnly`.
- `ActionsLendNamespace.getPositions(walletAddress, params?)` and `WalletLendNamespace.getPositions(params?)` (uses `wallet.address` implicitly).

### CLI
- `actions wallet lend positions [--chain <name> | --chain-id <id>] [--non-zero-only]` — one `wallet.lend.getPositions()` call; emits `LendMarketPosition[]` verbatim (bigints stringified). New `lendPositions` output sink.

### Demo
- Backend: `getLendPositions` service + `GET /wallet/lend/positions` route/controller (optional `chainId`/`nonZeroOnly` query).
- Frontend: `actionsApi.getPositions`, `EarnOperations.getPositions` on both wallet operation builders.
- `useLendProvider.ts`: both fan-out sites (mount load + post-trade refresh) now call `operations.getPositions()` once. The single `getPosition` activity-log entry is now honest (one call, one line).

## Testing
- SDK: provider walk / empty allowlist / missing-address / chain validation / per-market failure isolation; namespace aggregation / provider filter / `nonZeroOnly`; wallet implicit-address path.
- CLI: emit shape, `--chain` + `--non-zero-only` forwarding, multi-chain rejection, missing-lend/RPC error mapping.
- Backend: service param-forwarding + bigint serialization + error propagation.
- All suites green (SDK 673, CLI 267, backend 132, frontend 73); typecheck + lint clean across packages.
- Manual CLI run against the demo config (Base Sepolia + OP Sepolia): default aggregates both chains; `--non-zero-only` → `[]`; `--chain base-sepolia` filters; both-flags rejected with a validation error.


packages/cli/src/commands/wallet/lend/positions.ts
```diff
@@ -0,0 +1,43 @@
...
+      { chainIds },
+    )
+  }
+  const chainId = chainIds?.[0]
+  try {
+    const positions = await wallet.lend.getPositions({
+      chainId,
+      nonZeroOnly: flags.nonZeroOnly,
+    })
+    printOutput('lendPositions', positions)
```
its-everdred: This only supports fetching positions for the first chain: I'd prefer if this gave the option for either a single chain ID or an array of chain IDs, or, if none provided, have it check all supported chains from the ActionsConfig. That would be:
1. an SDK change
2. a change here

its-everdred: > option for either a single chain ID or an array of chain IDs, or, if none provided, have it check all supported chains from the ActionsConfig

Addressed: the SDK now accepts mutually exclusive `chainId` or `chainIds` filters and defaults to all configured chains, while the CLI forwards single or comma-separated chain selections without truncating them.

packages/cli/src/commands/wallet/lend/positions.ts
```diff
@@ -0,0 +1,36 @@
...
+  const { wallet, config } = await walletContext()
+  requireLendCapability(wallet)
+  const chainIds = resolveChainFlags(
+    flags,
+    config.chains.map((c) => c.chainId),
+  )
+  try {
+    const positions = await wallet.lend.getPositions(
+      chainIds
+        ? { chainIds, nonZeroOnly: flags.nonZeroOnly }
```
its-everdred: id rather a top level `options` with `nonZeroOnly` as the attr within it

packages/demo/backend/src/controllers/wallet.ts
```diff
@@ -94,6 +104,36 @@ export class WalletController {
...
+
+    const positions = await walletService.getLendPositions({
+      wallet,
+      params: {
+        chainId: chainId ? (Number(chainId) as SupportedChainId) : undefined,
+        nonZeroOnly: nonZeroOnly === 'true',
+      },
+    })
+    return c.json({ result: positions })
+  }
```
its-everdred, line 144: We would also want this to support passing multiple chain IDs as well.

packages/demo/frontend/src/api/actionsApi.ts
```diff
@@ -105,6 +105,28 @@ class ActionsApiClient extends BaseApiClient {
     }
   }
 
+  async getPositions(
+    params: { chainId?: SupportedChainId; nonZeroOnly?: boolean } = {},
+    headers: HeadersInit = {},
+  ): Promise<LendMarketPosition[]> {
```
its-everdred, line 114: We would want to support passing multiple chain IDs at this level, as well as all other levels of the SDK, the demo, and the CLI.

packages/demo/frontend/src/components/earn/frontendWalletOperations.ts
its-everdred, line 92: im surprised i dont see borrow actions here, do we support borrow actions with frontend wallets? respond to this comment, don't implement changes

packages/demo/frontend/src/components/earn/serverWalletOperations.ts
its-everdred, line 38: im surprised i dont see borrow actions here, where do we support borrow actions with server wallets? respond to this comment.

packages/cli/src/commands/wallet/lend/positions.ts
its-everdred: > top level `options` with `nonZeroOnly` as the attr within it

Addressed: `nonZeroOnly` now lives at `params.options.nonZeroOnly` throughout the SDK, CLI, and demo adapters.

packages/demo/frontend/src/hooks/__tests__/useLendProviderPositions.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+      )
+      return match ?? position(marketId.address, 0n)
+    }),
+    getPositions: vi.fn().mockResolvedValue(positions),
+    mintAsset: vi.fn().mockResolvedValue({}),
+    openPosition: vi.fn().mockResolvedValue({}),
+    closePosition: vi.fn().mockResolvedValue({}),
+    executeSwap: vi.fn().mockResolvedValue({}),
+  }
+}
```
its-everdred: do we have existing mocks or test utils for these? otherwise lets move to a new test helpers, utils, or mocks file and import

packages/demo/frontend/src/components/earn/frontendWalletOperations.ts
its-everdred: > do we support borrow actions with frontend wallets?

Yes. `buildFrontendBorrowOperations` is lower in this file and wires frontend-wallet position reads, quotes, open/close, collateral deposit/withdrawal, and repayment; this hunk is only the lend-operation builder.

packages/demo/frontend/src/hooks/useLendProvider.ts
```diff
@@ -93,51 +176,18 @@ export function useLendProvider({
...
-          isLoadingPosition: false,
-          marketId: market.marketId,
-          provider: market.provider,
-        })),
-    )
+    // One SDK call aggregates every market/provider position (see #14),
+    // replacing the per-market getPosition fan-out. Kept best-effort: the
+    // event-driven caller invokes this as `void`, so a refresh failure must
+    // log rather than surface as an unhandled rejection (the old per-market
+    // fan-out swallowed errors the same way).
```
its-everdred: Find all multi-line comments that we added and reduce them down to a more concise single-line comment in this PR.

packages/demo/frontend/src/hooks/useLendProvider.ts
```diff
@@ -163,82 +213,28 @@ export function useLendProvider({
...
-              ],
-              result.position,
-            )
-          }
-        }
+        seedPositionCache(queryClient, positions)
 
         // Build initial market positions array with all markets that have deposits
-        const initialPositions = positionResults
-          .filter((result) => {
```
its-everdred, line 203: DRY these lines, we do this twice

packages/sdk/src/actions/lend/core/LendProvider.ts
```diff
@@ -183,6 +185,48 @@ export abstract class LendProvider<
...
+    )
+
+    const settled = await Promise.allSettled(
+      markets.map((market) =>
+        this.getPosition(walletAddress, {
+          address: market.address,
+          chainId: market.chainId,
+        }),
+      ),
+    )
```
its-everdred, line 213: Instead of settling it per provider, I wonder if it makes more sense to set it across all providers at a higher level up. Respond with a comment with a few code snippets that would show off what this change could look like.

packages/demo/backend/src/controllers/wallet.ts
its-everdred: Addressed in `a4a1f5ee`: the backend accepts mutually exclusive `chainId` or comma-separated `chainIds` filters and forwards the parsed list to the SDK.

packages/demo/frontend/src/api/actionsApi.ts
its-everdred: Addressed in `a4a1f5ee`: `chainIds` now flows through the frontend API and backend, completing the existing SDK and CLI support.

packages/demo/frontend/src/hooks/__tests__/useLendProviderPositions.spec.ts
its-everdred: Addressed in `d0ca18c6`: moved the lend fixtures and hook wrapper to `src/test-utils/lendFixtures.ts`.

packages/sdk/src/actions/lend/namespaces/ActionsLendNamespace.ts
```diff
@@ -1,7 +1,31 @@
...
+   * @param params - Optional chain/provider filters and zero-balance toggle
+   * @returns Promise resolving to the wallet's positions
+   */
+  async getPositions(
+    walletAddress: Address,
+    params: GetPositionsParams = {},
+  ): Promise<LendMarketPosition[]> {
+    return this.fetchPositions(walletAddress, params)
+  }
+}
```
its-everdred: Is there any reason this shouldn't live in BaseLendNamespace? just rename that fetchPositions to getPositions and we can remove the changes in this file completely

packages/sdk/src/actions/lend/namespaces/WalletLendNamespace.ts
```diff
@@ -63,6 +64,21 @@ export class WalletLendNamespace extends BaseLendNamespace {
...
+   * every configured provider's market allowlist using `this.wallet.address`,
+   * isolates per-market RPC failures, and returns the same `LendMarketPosition`
+   * shape `getPosition` returns.
+   * @param params - Optional chain/provider filters and zero-balance toggle
+   * @returns Promise resolving to the wallet's positions
+   */
+  async getPositions(
+    params: GetPositionsParams = {},
+  ): Promise<LendMarketPosition[]> {
+    return this.fetchPositions(this.wallet.address, params)
```
its-everdred: again, id rather the BaseLendNamespace function be getPositions

packages/sdk/src/actions/lend/core/LendProvider.ts
its-everdred: That can move into `BaseLendNamespace`, but a direct move changes failure granularity. The simple provider-level version would be:

```ts
// LendProvider
return Promise.all(
  markets.map((market) => this.getPosition(walletAddress, market)),
)
```

```ts
// BaseLendNamespace
const settled = await Promise.allSettled(
  providers.map((provider) => provider.getPositions(walletAddress, params)),
)
const positions = settled.flatMap((result) =>
  result.status === 'fulfilled' ? result.value : [],
)
```

That drops every result from a provider when one of its markets fails. To retain per-market partial results with one higher-level settle point, the provider could expose its selected market IDs and the namespace could flatten all reads first:

```ts
const reads = providers.flatMap((provider) =>
  provider.getPositionMarketIds(params).map((marketId) =>
    provider.getPosition(walletAddress, marketId),
  ),
)
const settled = await Promise.allSettled(reads)
```

I would use the flattened version if we move this, since it centralizes settlement without weakening per-market isolation.

packages/demo/frontend/src/hooks/useLendProvider.ts
its-everdred: Addressed in `677cc44b`: PR-added prose blocks are now one line or removed; required public API JSDoc remains concise and tag-structured.

packages/demo/frontend/src/hooks/useLendProvider.ts
its-everdred: Addressed in `ee94e1c6`: mount and refresh now share `loadMarketPositions` for fetching, cache seeding, and projection.

packages/sdk/src/actions/lend/namespaces/ActionsLendNamespace.ts
its-everdred: Addressed in `15b56b86`: `BaseLendNamespace` now owns public `getPositions`, and `ActionsLendNamespace.ts` is unchanged from `main`.

packages/sdk/src/actions/lend/namespaces/WalletLendNamespace.ts
its-everdred: Addressed in `15b56b86`: the wallet namespace inherits the typed `getPositions(params?)` method and supplies its address lazily, so smart-wallet construction remains valid.

packages/sdk/src/actions/lend/namespaces/WalletLendNamespace.ts
```diff
@@ -17,12 +20,12 @@ import type { Wallet } from '@/wallet/core/wallets/abstract/Wallet.js'
...
  */
-export class WalletLendNamespace extends BaseLendNamespace {
+export class WalletLendNamespace extends BaseLendNamespace<WalletGetPositionsArgs> {
   constructor(
     providers: LendProviders,
     private readonly wallet: Wallet,
   ) {
-    super(providers)
+    super(providers, () => wallet.address)
   }
```
its-everdred, line 27: curious why we needed this change?

its-everdred: Because `getPositions` now lives on `BaseLendNamespace`, the base must preserve both public signatures: `actions.lend.getPositions(walletAddress, params?)` and `wallet.lend.getPositions(params?)`. `WalletGetPositionsArgs` keeps the wallet method typed without an address argument, while the callback supplies the bound wallet address. The callback must be lazy because smart-wallet namespaces are constructed before the address is initialized; passing `wallet.address` directly caused construction to throw `Smart wallet not initialized`.

its-everdred: It's odd to see a type that is specific to a single function, even though this class has many and relates to many. Perhaps we could change the type name. And add a comment that explains why we do () => wallet.address

its-everdred: Addressed in `88f62d0a`: the class generic now represents the namespace scope (`'actions' | 'wallet'`) instead of naming one function's arguments. I also added a concise comment explaining that the callback defers the address read until after smart-wallet namespace construction.

packages/sdk/src/actions/lend/namespaces/BaseLendNamespace.ts
```diff
@@ -45,6 +61,33 @@ export abstract class BaseLendNamespace extends BaseNamespace<
...
+   * @param args - Wallet address and filters, or wallet-scoped filters
+   * @returns Promise resolving to the wallet's positions across providers
+   * @throws AddressRequiredError, InvalidParamsError, or ChainNotSupportedError
+   */
+  async getPositions(
+    ...args: GetPositionsArgs<TScope>
+  ): Promise<LendMarketPosition[]> {
+    const { walletAddress, params } = this.resolveGetPositionsArgs(args)
+    const providers = params.provider
+      ? [this.providers[params.provider]].filter(
```
its-everdred: add a comment with example

packages/sdk/src/actions/lend/namespaces/BaseLendNamespace.ts
```diff
@@ -45,6 +61,33 @@ export abstract class BaseLendNamespace extends BaseNamespace<
...
+          (provider): provider is ConfiguredLendProvider =>
+            provider !== undefined,
```
its-everdred: do we have a getProvider() function? if not, make one

packages/sdk/src/actions/lend/namespaces/BaseLendNamespace.ts
```diff
@@ -65,4 +108,19 @@ export abstract class BaseLendNamespace extends BaseNamespace<
       details: `No provider configured for market on chain ${marketId.chainId}`,
     })
   }
+
+  private resolveGetPositionsArgs(args: GetPositionsArgs<LendNamespaceScope>): {
+    walletAddress: Address
+    params: GetPositionsParams
+  } {
+    if (typeof args[0] === 'string') {
```
its-everdred: I dont like "args", Follow patterns established in the repo. Call this `params` and use named param attrs. Let's not sort by whether or not it's a string.

its-everdred: include wallet address as a param

packages/sdk/src/actions/lend/namespaces/BaseLendNamespace.ts
its-everdred: Addressed in `e3f15d27`: the JSDoc now includes named-parameter examples for both `actions.lend` and `wallet.lend`.

packages/sdk/src/actions/lend/namespaces/BaseLendNamespace.ts
its-everdred: Addressed in `ea1528ae`: `BaseNamespace.getProvider(name)` now returns configured providers and throws `ProviderNotConfiguredError` when an explicit provider is unavailable; lend aggregation uses it.

packages/sdk/src/actions/lend/namespaces/BaseLendNamespace.ts
its-everdred: Addressed in `e3f15d27`: `getPositions` now accepts one named `params` object containing `walletAddress` and filters. The string-discrimination path is removed, and wallet namespaces fill their bound address when it is omitted.

jefr90 (approved):
LGTM

PR #512, opened 2026-06-18
Fix earn demo borrow/withdraw/swap issues

its-everdred:
Fixes for the Aave/Uniswap earn demo (`packages/demo/frontend`). 112 tests, lint, build green. No backend changes.

- **Repay-max revert + raw error leak** — repaying max USDC reverted with "ERC20: transfer amount exceeds balance" (the repay gate read the cross-chain total, but USDC_DEMO sits on both Base + OP Sepolia, so `max` prefilled more than the borrow chain could spend), and the raw revert calldata leaked into the modal. Fix: scope the gate to the market's chain; show `ActionsError.shortMessage`.
- **Health bar overshoot** — after a successful borrow the utilization bar shot to ~100% before settling to ~60%, because the optimistic position update landed (debt now includes the borrow) while the form amount was still set, double-counting it. Fix: zero the projected delta while the tx executes.
- **Withdraw "You withdraw 0"** — the withdraw review modal showed 0 when withdrawing 0.0001, because it rendered the raw `amount` state (empty when the illiquid Aave amount is locked) instead of the effective amount. Fix: use `effectiveAmount`, and feed it to the health projection too.
- **Testnet withdraw error** — illiquid Aave ETH withdrawals can revert on low testnet liquidity but only showed "Transaction failed". Fix: surface a testnet-specific message for that market+mode.
- **Balance flicker $70→$140→$70** — switching to the Borrow tab auto-pledges the lent Morpho shares; the pledged collateral landed (optimistic) before the direct lend shares refetched to zero, so the same shares were summed twice. Fix: the reconcile moves direct→pledged in one optimistic update, and `buildEffectiveLendPositions` reads it as a fallback so the total never double-counts.
- **Swap revert** — USDC→OP Uniswap swaps sometimes reverted with only "tx failed". Pool/fee/liquidity check out, so it's a runtime revert needing an on-chain repro; pinned pending one. The swap now surfaces the real reason instead of "tx failed".

Also: `mise.toml` `[alias]`→`[tool_alias]`.


its-everdred (commented):
reduce all multi-line comments we added in this PR to single line

packages/demo/frontend/src/components/earn/borrow/ReviewBorrowHealthModal.tsx
```diff
@@ -89,6 +89,9 @@ export function ReviewBorrowHealthModal({
 }: ReviewBorrowHealthModalProps) {
   const symbol = displaySymbol(asset.metadata.symbol)
   const assetLogo = getAssetLogo(asset.metadata.symbol)
+  // Format the (possibly full-precision) amount as a label: first two decimals
+  // full size, the rest dimmed and smaller.
```
its-everdred:
```suggestion
```

its-everdred:
> reduce all multi-line comments we added in this PR to single line

Done in `9d218695`. I collapsed the added multi-line comments into single-line comments, preserved your latest deletion in `ReviewBorrowHealthModal.tsx`, and removed the added em dash in a comment.

mise.toml
```diff
@@ -12,7 +12,7 @@ anvil = "nightly-0dd4d3153764f4706c2c9857675e42dec64155a7"
 # Supersim
 supersim = "0.1.0-alpha.58"
 
-[alias]
+[tool_alias]
```
its-everdred, line 15: `[alias]` was deprecated for `[tool_alias]`

its-everdred:
> i meant greatly reduce the comments themselves to a short phrase if needed, otherwise delete

Done in `f84e2ca0`. I deleted the comments that only restated nearby code and reduced the remaining ones to short context phrases.

mise.toml
its-everdred: > `[alias]` was deprecated for `[tool_alias]`

Already addressed on this branch: `mise.toml` now uses `[tool_alias]`, and the PR diff shows `[alias]` was replaced.

jefr90 (approved):
LGTM

PR #524, opened 2026-07-07
fix: encode Uniswap V4 recipients

its-everdred:
## Problem
Uniswap V4 swaps can send output tokens to a recipient chosen by the caller. The previous router encoding computed that recipient, but the final withdrawal action ignored it. As a result, swaps with a custom recipient could complete while sending the output to the wrong address.

## Solution
Encode the V4 withdrawal as a recipient-aware `TAKE` action so exact-in and exact-out swaps pay the resolved output recipient.

- The swap action still carries the slippage bound.
- The withdrawal now uses the full open output delta for the requested recipient.
- Regression coverage decodes the router action bytes and `TAKE` parameters.

## Related
- https://github.com/ethereum-optimism/actions/pull/500#discussion_r3484766092

### DIFF
--- a/.changeset/uniswap-recipient-take.md
+++ b/.changeset/uniswap-recipient-take.md
@@ -1,0 +1,5 @@
+---
+"@eth-optimism/actions-sdk": patch
+---
+
+Fix Uniswap universal router swaps to encode custom output recipients in V4 take actions.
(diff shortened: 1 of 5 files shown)


jefr90 (approved):
LGTM

PR #541, opened 2026-07-15
Demo: Privy wallet provider fix

its-everdred:
# Problem

Privy signups blocked: the app's Privy instance hit its dev-mode user limit (`max_accounts_reached`) so new users can't sign in, and the failure was invisible because login errors were silently swallowed.

# Solution

- Surface Privy login errors (log + clear message + retry) instead of a silent redirect that stripped the wallet-provider param and bounced users back to the picker.
- Add enforced Netlify security headers (`X-Frame-Options: DENY`, nosniff, Referrer-Policy, HSTS, and an allow-by-default `Content-Security-Policy` that locks framing/plugins/base-uri) to satisfy Privy's production "Secure your app" checklist and unblock the move to prod.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -6,7 +6,11 @@
   NODE_VERSION = "20"
   NODE_OPTIONS = "--max-old-space-size=8192"
 
-# Security headers for moving the Privy app to production; connect-src origins are grouped by service.
+# Security headers for moving the Privy app to production. The CSP is allow-by-default
+# so every third-party origin the demo uses (Privy, Dynamic, Turnkey, WalletConnect,
+# fonts, analytics, RPCs) keeps working without enumeration; it only locks what matters:
+# nothing can iframe the wallet (frame-ancestors), no plugins, no <base> hijacking.
+# Trade-off: 'unsafe-inline'/'unsafe-eval' mean this is not XSS-hardened.
 [[headers]]
   for = "/*"
   [headers.values]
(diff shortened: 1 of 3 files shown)


netlify.toml
```diff
@@ -6,3 +6,17 @@
...
+    X-Frame-Options = "DENY"
+    X-Content-Type-Options = "nosniff"
+    Referrer-Policy = "strict-origin-when-cross-origin"
+    Strict-Transport-Security = "max-age=31536000; includeSubDomains"
+    # Allow-by-default: every origin the demo uses (Privy, Dynamic, Turnkey,
+    # WalletConnect, fonts, analytics, RPCs) works without enumeration, and only the
+    # dangerous things are locked: nothing can iframe the wallet (frame-ancestors),
+    # no plugins (object-src), no <base> hijacking. Trade-off: 'unsafe-inline' and
+    # 'unsafe-eval' mean it is not XSS-hardened.
+    Content-Security-Policy = "default-src 'self' https: wss: data: blob: 'unsafe-inline' 'unsafe-eval'; frame-ancestors 'none'; object-src 'none'; base-uri 'self'"
```
its-everdred, line 22: Allow-by-default Content-Security-Policy for the whole site: it permits every origin the app uses (so there's no per-service allowlist to maintain) and only locks the risky parts. `frame-ancestors 'none'` stops any other site from iframing the app, which protects the embedded Privy wallet from clickjacking; `object-src 'none'` blocks plugins; `base-uri 'self'` blocks `<base>` tag hijacking. Required to move the Privy app to production.

netlify.toml
```diff
@@ -6,3 +6,17 @@
   NODE_VERSION = "20"
   NODE_OPTIONS = "--max-old-space-size=8192"
 
+# Security headers for moving the Privy app to production.
```
its-everdred:
```suggestion
# Security headers needed for Privy in Prod mode
```

tremarkley (approved):

jefr90 (approved):

PR #545, opened 2026-07-21
Prevent Turnkey refresh flashes

its-everdred:
# Problem

This fix is now part of the combined Turnkey authentication PR. Keeping both PRs open would duplicate the refresh changes.

# Solution

Superseded by #551.


its-everdred:
Combined into #551.

PR #546, opened 2026-07-21
Show small ETH lend positions

its-everdred:
# Problem

Users who lend small ETH amounts cannot see their position after a refresh. Two-decimal flooring turns `0.001 ETH` into zero before empty rows are filtered.

# Solution

Use four-decimal downward precision for ETH positions.


jefr90 (approved):
LGTM

PR #547, opened 2026-07-21
Fix Dynamic login modal width

its-everdred:
# Problem

Users see a broken Dynamic login modal with empty space and a displaced close button. The demo wrapper is wider than the upgraded embedded widget.

# Solution

Shrink-wrap the wrapper to the widget with a responsive width cap.


jefr90 (approved):
LGTM

PR #548, opened 2026-07-21
Disable analytics in local development

its-everdred:
# Problem

Users running the demo locally see repeated Mixpanel CORS warnings in Firefox. A configured token sends analytics that Enhanced Tracking Protection blocks.

# Solution

Send Mixpanel analytics only from production builds.


jefr90 (approved):
LGTM

PR #549, opened 2026-07-21
Restore provider-neutral ETH faucet

its-everdred:
# Problem

Dynamic and Turnkey users cannot mint ETH because [#522](https://github.com/ethereum-optimism/actions/pull/522) made the faucet require a Privy session. Those providers authenticate outside Privy, so otherwise valid requests receive 401 Unauthorized.

# Solution

Accept a validated recipient address while retaining faucet protections.


packages/demo/backend/src/router.ts
```diff
@@ -6,29 +7,30 @@ import { fileURLToPath } from 'url'
...
-    keyGenerator: (c) => c.get('auth').rateLimitKey,
-  })
+const faucetRateLimit = rateLimiter<RouterEnv>({
+  windowMs: FAUCET_RATE_LIMIT_WINDOW_MS,
+  limit: FAUCET_RATE_LIMIT_MAX,
+  message: { error: 'Too many requests' },
+  keyGenerator: (c) => {
+    const address = getConnInfo(c).remote.address ?? 'unknown'
+    return `faucet-connection:${address}`
+  },
```
its-everdred, line 32: This replaces the Privy user key with the request’s network address, preserving faucet rate limiting without requiring provider-specific authentication.

jefr90 (approved):
LGTM

PR #551, opened 2026-07-21
Fix Turnkey login and refresh

its-everdred:
# Problem

Turnkey users see authentication screens flash during refresh. Passkey login starts without a user gesture, and the UI renders before wallet restoration completes.

# Solution

Require explicit passkey actions and wait for restored wallet state.

### DIFF
--- a/packages/demo/frontend/src/components/earn/EarnWithTurnkeyWallet.tsx
+++ b/packages/demo/frontend/src/components/earn/EarnWithTurnkeyWallet.tsx
@@ -3,17 +3,19 @@ import { AuthState, ClientState, useTurnkey } from '@turnkey/react-wallet-kit'
 import { EarnWithFrontendWallet } from './EarnWithFrontendWallet'
 import { WALLET_PROVIDERS } from '@/constants/walletProviders'
 import { LoginWithTurnkey } from './LoginWithTurnkey'
+import { WalletLoadingScreen } from './WalletLoadingScreen'
 import { useTurnkeyWallet } from '@/hooks/useTurnkeyWallet'
 import { trackEvent, identifyUser } from '@/utils/analytics'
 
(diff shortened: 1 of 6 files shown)


packages/demo/frontend/src/components/earn/EarnWithTurnkeyWallet.tsx
```diff
@@ -3,17 +3,19 @@ import { AuthState, ClientState, useTurnkey } from '@turnkey/react-wallet-kit'
...
   const { smartWallet } = useTurnkeyWallet()
-  const { clientState, authState, user, logout } = useTurnkey()
+  const { clientState, authState, session, user, logout } = useTurnkey()
 
   const isLoggedIn =
     clientState === ClientState.Ready &&
     authState === AuthState.Authenticated &&
-    user
+    Boolean(session) &&
+    Boolean(user)
```
its-everdred, line 18:
```suggestion
   session &&
   user
```

packages/demo/frontend/src/components/earn/EarnWithTurnkeyWallet.tsx
```diff
@@ -28,6 +30,14 @@ export function EarnWithTurnkeyWallet() {
     }
   }, [isLoggedIn, user])
 
+  const isRestoringSession =
+    clientState !== ClientState.Ready ||
+    Boolean(session) !== (authState === AuthState.Authenticated)
```
its-everdred, line 35:
```suggestion
   session !== (authState === AuthState.Authenticated)
```

its-everdred:
Closing: the Turnkey passkey-on-load error and refresh flash don't reproduce on current main, so this fix guards an issue that no longer exists. Reopen if it resurfaces.

PR #530, opened 2026-07-10
chore: automate dependency updates

its-everdred:
# Problem

The repository has no routine Dependabot schedule, so dependency updates can accumulate into large manual refreshes. The current lockfile also resolves many packages with known advisories, which would make #486 pin stale vulnerable versions.

# Solution

Refresh dependency resolutions to patched versions and schedule weekly update PRs.

- Group routine minor and patch updates while keeping major updates separate.
- Preserve published SDK peer ranges while pinning the tested Turnkey development matrix.
- Leave only upstream advisories without compatible fixes: `bigint-buffer` and `elliptic` have no patched release, while UUID v8/v9 requires parent migrations to v11.

Fixes #529.
Related: #486.

PR #552, opened 2026-07-21
Fix Aave reserve reads

its-everdred:
# Problem

Users cannot load Aave markets on Optimism Sepolia because reserve reads overflow. The Aave helper upgrade in [#530](https://github.com/ethereum-optimism/actions/pull/530) moved from 1.36.2 to 1.38.0, whose ABI cannot decode the older UI data provider still deployed there ([More details](https://gist.github.com/its-everdred/c6b71564f3195bb355c28eb4d3f93d6d)). We use Aave's packages only to aggregate and format lend-market reserve data while the rest of the provider already uses viem; version 1.38 retargeted the same `getReservesHumanized` helper to the V3.6+ tuple without a legacy or versioned method, so there is no newer SDK function compatible with Optimism Sepolia's older deployment.

# Solution

Read pool and reserve data directly with viem.

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -56,13 +56,10 @@
   "author": "Optimism PBC",
   "license": "MIT",
   "dependencies": {
-    "@aave/contract-helpers": "^1.30.0",
-    "@aave/math-utils": "^1.30.0",
     "@eth-optimism/viem": "^0.4.13",
     "@morpho-org/blue-sdk": ">=4.13.1 <4.14.0",
     "@morpho-org/blue-sdk-viem": ">=3.2.0 <3.3.0",
     "@morpho-org/morpho-ts": ">=2.4.6 <2.5.0",
-    "ethers": "^5.7.2",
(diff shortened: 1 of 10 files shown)


packages/sdk/package.json
```diff
@@ -56,13 +56,10 @@
   "author": "Optimism PBC",
   "license": "MIT",
   "dependencies": {
-    "@aave/contract-helpers": "^1.30.0",
-    "@aave/math-utils": "^1.30.0",
     "@eth-optimism/viem": "^0.4.13",
     "@morpho-org/blue-sdk": ">=4.13.1 <4.14.0",
     "@morpho-org/blue-sdk-viem": ">=3.2.0 <3.3.0",
     "@morpho-org/morpho-ts": ">=2.4.6 <2.5.0",
-    "ethers": "^5.7.2",
```
its-everdred, line 65: big win, this was a heavy dependency only used for an aave read function.

jefr90 (approved):
LGTM

PR #553, opened 2026-07-21
fix: Privy UX issues

its-everdred:
| Problem | Solution |
| --- | --- |
| New Privy users cannot delegate session signers because wallets created by the new local app use Privy's TEE authorization path, which calls `Buffer` in the browser; existing production wallets did not exercise this path. | Initialize the existing `buffer` dependency before the app loads. |
| Privy users can remain stuck on loading because signer setup starts before the embedded-wallet proxy is ready. | Wait for `useWallets().ready` before adding the session signer. |
| Privy users see “Unable to sign in” after intentionally closing the auth dialog because [#541](https://github.com/ethereum-optimism/actions/pull/541) treats `exited_auth_flow` as a failure. | Return cancellations to the wallet picker while preserving the error UI for genuine failures. |


jefr90 (approved):
LGTM

PR #465, opened 2026-05-11
Borrow PR #4: demo backend /borrow endpoints

its-everdred:
## Problem

The demo backend had no borrow HTTP routes, so PR #5's frontend could not drive borrow flows. The PR also had no shared backend-side error mapping, no branded request schemas, and no controller-level coverage.

## Solution

- **Borrow routes**
  - add nine `/borrow/*` and `/wallet/borrow/*` endpoints that thinly proxy the SDK `BorrowProvider`

  ```
  GET   /borrow/markets                                       public
  POST  /borrow/price                                         public
  POST  /borrow/quote                                         auth
  GET   /wallet/borrow/:chainId/:marketId/position            auth

  POST  /borrow/position/open                                 auth
  POST  /borrow/position/close                                auth
  POST  /borrow/position/deposit-collateral                   auth
  POST  /borrow/position/withdraw-collateral                  auth
  POST  /borrow/position/repay                                auth
  ```

  - mutation bodies accept fresh params or a pre-built `{ quote: BorrowQuote }` discriminated on `action`
  - `/borrow/quote` derives `walletAddress` from the auth idToken; body schema is strict so a caller-supplied `walletAddress` returns 400
  - mutation receipts decorated with `blockExplorerUrls` server-side

- **Validation and error handling**
  - branded zod schemas in `helpers/schemas.ts` (`AddressSchema`, `Bytes32Schema`, `ChainIdSchema`, `ChainIdStringSchema`, `AmountExactSchema`, `AmountWithMaxSchema`, `BorrowMarketIdSchema`)
  - shared `mapSdkError` table covers every SDK `ActionsError` subclass plus a local `WalletNotFoundError`; meta-coverage test fails CI when a new SDK class isn't mapped or allowlisted
  - borrow-scoped `app.onError` global handler replaces per-route try/catch on borrow routes
  - `validateRequest` returns `{ error, details: Array<{ path, message }> }` instead of raw zod issues, rejects malformed JSON with a structured 400, and routes through the same envelope as `mapSdkError`

- **Backend wiring**
  - `MorphoBorrowDemo` `BorrowMarketConfig` literal mirrors `morpho.borrow` in `deployments.json` for chain 84532
  - `actions.borrow` configured via `borrow: { morpho: { marketAllowlist: ALL_BORROW_MARKETS } }`
  - `WalletController.getBorrowPosition` routes the read path through the SDK's `actions.borrow.getPosition`
[...]

PR #554, opened 2026-07-22
fix: getPosition() sometimes fails

its-everdred:
| Problem | Solution |
| --- | --- |
| Users cannot load lending positions, borrow markets, or swap quotes because bodyless GETs advertise JSON, causing the backend to parse an empty body and return `400 Invalid JSON body`. [PR #465](https://github.com/ethereum-optimism/actions/pull/465) made advertised JSON parse failures fatal, and [PR #508](https://github.com/ethereum-optimism/actions/pull/508) later shipped the positions endpoint with this mismatch. | Add the JSON content type only when a request has a body, and cover authenticated position reads with a regression test. |

### DIFF
--- a/packages/demo/frontend/src/api/actionsApi.spec.ts
+++ b/packages/demo/frontend/src/api/actionsApi.spec.ts
@@ -89,4 +89,27 @@ describe('ActionsApiClient', () => {
       }),
     )
   })
+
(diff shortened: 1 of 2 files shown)


its-everdred:
Combined into #555 for final demo QA.

tremarkley (approved):

PR #311, opened 2026-03-13
Add support for Aero/Velodrome

its-everdred:
## Summary

Adds **Velodrome (Optimism) and Aerodrome (Base)** swap support, plus a major refactor of the swap interface for multi-provider quoting.

velodrome-demo

---

### Velodrome/Aerodrome Provider

- **V2 (AMM) pools** and **CL (Slipstream) pools** via QuoterV2
- **Router support**: Universal, Leaf, and standard V2 routers
- **12 chains**: OP Mainnet, Base, Mode, Lisk, Fraxtal, Metal, Soneium, World, Bob, Ink, Swell, Unichain
- **`{ contracts, metadata }` address pattern** — reference impl for #328
- **Encoding split** into `v2.ts`, `cl.ts`, `helpers.ts` submodules
- **Full test coverage**: unit, routing, encoding, and Anvil fork tests

### Swap Interface Refactor

- **`getQuote()` replaces `getPrice()`** — flat `SwapQuote` with `Raw` bigint suffixes (`amountInRaw`, `amountOutRaw`, `amountOutMinRaw`)
- **`getQuotes()`** — all provider quotes sorted by best price
- **`getBestQuote()`** — best quote across providers via parallel fetching
- **`SwapSettings`** — shared config (slippage, routing, defaultProvider) with provider-level override precedence
- **`_execute` → `_getQuote` + `_executeFromQuote`** — no duplicated quoting/encoding

### SDK Infrastructure

- **Provider type registry** — centralized provider map, all wallet providers updated
- **EIP-55 address validation** — `validateAddressMap()` / `validateAssetAddresses()` with full test suite
- **New chains**: Ink, Soneium, Mode added to supported chains and assets
- **Aave/Morpho** refactored to unified address config format
- **Replaced local ERC20 ABIs** with viem `erc20Abi`

### Demo App

- New `/swap/quote` endpoint, provider param threaded through swap routes
- `SwapMarketSelector`, `DemoProviderTooltip`, `MarketOption`, shared `Dropdown` components
- Chain and provider logos, UI bug fixes
- Foundry deploy script for Velodrome test markets

### CI/CD

- Docker build migrated from CircleCI to GitHub Actions

---

## Follow-up Issues

- **#328** — Standardize `{ contracts, metadata }` pattern across all providers
- **#331** — `includeCalldata` flag to skip calldata when wallet not needed
- **#332** — Add network fork tests to CI

PR #555, opened 2026-07-22
fix: resolve remaining demo regressions

its-everdred:
| Problem | Solution |
| --- | --- |
| Users cannot load positions, borrow markets, or swap quotes because bodyless GET requests advertise JSON and the backend rejects their empty bodies ([#465](https://github.com/ethereum-optimism/actions/pull/465), [#508](https://github.com/ethereum-optimism/actions/pull/508)). | Include the JSON content type only when a request has a body, with regression coverage from [#554](https://github.com/ethereum-optimism/actions/pull/554). |
| Users can trigger an endless stream of collateral-deposit requests when automatic Morpho reconciliation fails because [#490](https://github.com/ethereum-optimism/actions/pull/490) clears its attempt marker and [#512](https://github.com/ethereum-optimism/actions/pull/512) restores the same eligible position. | Retain the once-per-mount marker after failure and cover rejection-driven rerenders. |
| Users cannot auto-pledge Morpho collateral because the backend rejects the SDK-supported `{ max: true }` deposit payload before it reaches the provider. | Accept max amounts at the deposit-mutation boundary and cover the authenticated HTTP route. |
| Privy users see “Unable to sign in” after intentionally closing the auth dialog because [#541](https://github.com/ethereum-optimism/actions/pull/541) treats `exited_auth_flow` as a failure. | Return cancellations to the wallet picker while preserving the error UI for genuine failures. |
| Users cannot repay Aave debt despite holding USDC because [#512](https://github.com/ethereum-optimism/actions/pull/512) checks Base-only `USDC_DEMO` on the Aave market's OP Sepolia chain and always reads zero. | Gate mirrored Aave repayment on the Base balance while keeping ordinary repayment scoped to its market chain. |
| Users briefly lose every displayed balance and cannot swap after borrow or repay because [#311](https://github.com/ethereum-optimism/actions/pull/311) registered an empty fetcher for the shared balance query, which transaction refreshes can execute. | Keep Swap passive while sharing the real balance fetcher, with invalidation regression coverage. |
| Users see an infinite health factor for an active loan when the borrow or repay amount is zero because [#466](https://github.com/ethereum-optimism/actions/pull/466) treats the absence of a projection as a debt-free position. | Derive the fallback from current collateral and debt; reserve infinity for no debt. |
| Isolated provider fixes do not prove the full demo works when combined. | Temporarily stack [#551](https://github.com/ethereum-optimism/actions/pull/551), [#552](https://github.com/ethereum-optimism/actions/pull/552), and [#553](https://github.com/ethereum-optimism/actions/pull/553) for final UX QA; issues remain owned by those PRs and these commits will be removed here before merge. |

### DIFF
--- a/.changeset/fix-aave-reserve-reads.md
+++ b/.changeset/fix-aave-reserve-reads.md
@@ -1,0 +1,5 @@
+---
+'@eth-optimism/actions-sdk': patch
+---
+
+Read Aave lend reserves directly with viem to avoid UI data-provider ABI mismatches.
--- a/packages/demo/backend/src/controllers/borrow.routes.spec.ts
+++ b/packages/demo/backend/src/controllers/borrow.routes.spec.ts
@@ -62,8 +62,9 @@ vi.mock('@/middleware/actions.js', () => ({
 const MARKET_ID = {
   kind: 'morpho-blue' as const,
   chainId: 84532,
-  marketId: '0x' + 'a'.repeat(64),
-}
+  marketId:
+    '0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
+} as const
 const WALLET = '0xaabbccddeeff00112233445566778899aabbccdd'
 
 beforeEach(async () => {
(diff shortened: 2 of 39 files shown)


its-everdred:
Closing this catch-all. It existed only to QA every demo fix stacked together; QA has passed, so each fix has been split into its own context PR. No commits are lost, every fix below lives on its listed PR:

**Wallet providers**
- Turnkey login + refresh flashes → #551
- Privy `Buffer` polyfill, wait-for-wallets, canceled login → #553

**Aave**
- Aave reserve reads (OP Sepolia overflow) → #552
- Aave repayment balance + zero-input health factor → #557

**Morpho / borrow**
- Collateral retry loop + max collateral deposits → #556

**API / earn UX**
- Omit JSON headers from bodyless requests → #554
- Preserve balances during refresh + disable unfunded lend input → #558

PR #556, opened 2026-07-22
fix: stop Morpho collateral retry loop and accept max deposits

its-everdred:
| Problem | Solution |
| --- | --- |
| When a user opens the Borrow tab, the demo automatically pledges their existing collateral. If that automatic deposit fails even once, the app retries it on every re-render, firing an endless stream of failed `deposit-collateral` requests and leaving the user stuck. | Keep the once-per-mount "already attempted" marker set after a failure so a failed reconciliation is not retried on every render. Adds a regression test covering a rejected deposit followed by re-renders. |
| A user cannot auto-pledge their full collateral balance because the backend rejects the "deposit the maximum" payload (`{ max: true }`) before it reaches the lending provider. | Accept the max-amount payload at the deposit route and pass it through to the provider. Adds a test for the authenticated deposit route with a max amount. |

### DIFF
--- a/packages/demo/backend/src/controllers/borrow.ts
+++ b/packages/demo/backend/src/controllers/borrow.ts
@@ -128,7 +128,7 @@ export async function closePosition(c: Context) {
 
 const DepositCollateralParamsBody = z.strictObject({
   marketId: BorrowMarketIdSchema,
-  amount: AmountExactSchema,
+  amount: AmountWithMaxSchema,
 })
 
 const DepositCollateralRequestSchema = z.object({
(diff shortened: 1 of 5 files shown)


packages/demo/backend/src/controllers/borrow.ts
```diff
@@ -128,7 +128,7 @@ export async function closePosition(c: Context) {
 
 const DepositCollateralParamsBody = z.strictObject({
   marketId: BorrowMarketIdSchema,
-  amount: AmountExactSchema,
+  amount: AmountWithMaxSchema,
```
its-everdred, line 131: this allows for both exact value and also `{ max: true }`

tremarkley (approved):

PR #557, opened 2026-07-22
fix: restore Aave repayment balance and health factor

its-everdred:
| Problem | Solution |
| --- | --- |
| A user who holds USDC still cannot repay their Aave loan. The repay screen reads the balance on the wrong chain (it checks a Base-only demo token while the Aave market lives on OP Sepolia), always sees zero, and tells the user to "Get USDC" by swapping for demo tokens that can never repay the loan, leaving them stuck. | Read the repayment balance from the correct chain for the mirrored Aave demo market, while ordinary repayments keep reading the balance on their own market's chain. |
| A user with an active loan sees an "infinite" health factor whenever the borrow or repay input is zero, because an empty input is mistaken for a debt-free position. | When the input is zero, derive the health factor from the position's current collateral and debt; only show infinity when there is genuinely no debt. |

### DIFF
--- a/packages/demo/frontend/src/demoMagic/aaveDemoMagic.ts
+++ b/packages/demo/frontend/src/demoMagic/aaveDemoMagic.ts
@@ -23,12 +23,22 @@ export function isMirrorMarket(marketId: BorrowMarketId): boolean {
   return sameMarketId(marketId, AaveETHBorrowUSDCDemo)
 }
 
-/** Returns USDC_DEMO for the mirror market (user holds USDC_DEMO, not the real borrowed USDC); otherwise returns the real borrow asset. */
-export function repayGateAsset(
+/**
+ * Resolve the balance used to gate a repayment.
+ * @description Uses the Base USDC_DEMO mirror for the Aave demo and the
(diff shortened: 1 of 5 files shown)


packages/demo/frontend/src/demoMagic/aaveDemoMagic.ts
```diff
@@ -23,12 +23,22 @@ export function isMirrorMarket(marketId: BorrowMarketId): boolean {
...
+ * @returns Asset and chain containing the spendable gate balance
+ */
+export function resolveRepayBalanceSource(
   market: BorrowMarket | null,
   borrowAsset: Asset | null,
-): Asset | null {
-  return market && isMirrorMarket(market.marketId) ? USDC_DEMO : borrowAsset
+) {
+  if (market && isMirrorMarket(market.marketId)) {
+    return { asset: USDC_DEMO, chainId: baseSepolia.id }
```
tremarkley: nit: you could consider creating a `mirrorMarketChainId` variable and setting it to `baseSepolia`

tremarkley (approved):

PR #589, opened 2026-09-04
ci: run renovate every 4 hours

falcorocks:
Reduces the Renovate scan cadence from hourly to every 4 hours (staggered minutes preserved). Balances the advisory-to-fix loop against runner cost; security-fix behavior is otherwise unchanged.


falcorocks:
Withdrawn: existing hourly cadence stays as-is; the cadence change only applies to the repos still being onboarded.

PR #600, opened 2026-09-17
security: patch smol-toml

its-everdred:
# Problem
The lockfile resolved vulnerable `smol-toml` versions ([finding](https://github.com/ethereum-optimism/vulnerability-management/issues/1260)).

# Solution
Override `smol-toml` to its patched release and regenerate the lockfile.


jefr90 (approved):
LGTM

